PR #68, opened 2024-04-18
Humanize last update time

daniellemaxwell:
### Scope of changes

Humanizes last update time in transactions table.
Displays `Accept` and `Reject` for incoming transfers in transactions table and detail modal.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/26855852?key=2e278d2135132ccec1f0371f91e9c8c0

### Author checklist

### Reviewer(s) checklist


bbengfort (approved):
Looks great; thanks for making this happen! 

I will say, I do forget sometimes that we're not doing a more standard API backend with fully reactive front-end and that has meant we have to jump through hoops with the htmx events. Please feel free to push back on some of these ideas if you think that the data conversion should be done in the backend! Things like a search dropdown we can't do, but certainly we could have humanized the datetime in the Go template instead of in Javascript. 

Still, there are a lot of UI utilities in javascript that might not be in Go. So maybe we should think about a webpack system that will allow us to import utilities from javascript and make front-end development easier?

PR #70, opened 2024-04-18
Ensure welcome message remains hidden after close

daniellemaxwell:
### Scope of changes

Updates script to ensure the welcome message remains hidden after a user clicks the close button. 

Note: It was decided to display the message when the user logs in again so that users may easily locate the instructions on starting a transfer.

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/26865950?key=885ed7e1545101a9ec682fbe92276243

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/web/static/js/main.js
+++ b/pkg/web/static/js/main.js
@@ -1,3 +1,26 @@
 document.body.addEventListener('htmx:configRequest', function(e) {
   e.detail.headers['Accept'] = 'text/html'
 });
+
+const hideWelcomeMsg = 'hideWelcomeMsg'
+const welcomeMsg = document.getElementById('welcome-msg')
+const closeMsg = document.getElementById('close-welcome-msg')
+const logoutBttn = document.getElementById('logout-bttn')
+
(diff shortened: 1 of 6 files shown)


bbengfort (approved):
Oh boy, we're starting to get into front-end state management now! 

This looks great and is a very good start. 

In the future, we might want to make state management more generic; e.g. save state like this and other front-end state in a single larger object. We don't have to use something as heavy handed as Redux, but we could create our own lightweight state management library for configuration. But that is only possible if we have webpack or other front-end helpers. 

Maybe something like https://github.com/pmndrs/zustand?

daniellemaxwell:
@bbengfort Zustand is fine. I would prefer to use a collapsible component here. I tested the daisyUI component, but it's not collapsed by default and I didn't want to spend more time trying to figure out which class to override. I'll look into using Flowbite or another library.

PR #78, opened 2024-04-22
Display toast notification for response errors

daniellemaxwell:
### Scope of changes

Adds a toast element to display error response messages.

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/26973051?key=b3619bab743a7e2bebf888b90fa824d0

### Author checklist

### Reviewer(s) checklist


pkg/web/static/js/main.js
```diff
@@ -24,3 +24,21 @@ if (logoutBttn) {
     localStorage.removeItem(hideWelcomeMsg)
   })
 }
+
+document.body.addEventListener('htmx:responseError', (e) => {
+  // Close any open modals.
```
daniellemaxwell, line 29: When modals are open the toast appears behind it. This makes it easy for users to not notice them. I attempted to resolve this by setting the z-index really high, but that didn't work. We can return to this with follow on stories, as I wanted to ensure this gets added.

pkg/web/templates/layouts/base.html
```diff
@@ -9,6 +9,7 @@
 
   <main>
     {{ block "content" . }}{{ end }}
+    <div id="toast" class="toast toast-top toast-end"></div>
```
daniellemaxwell, line 12: `min-width: fit-content` is the default style applied to the toast elements. Should we update the error responses for the add counterparty modal? 

If not, we may have to amend the styling since attempting to override it with `!important` didn't work.

pkg/web/static/js/main.js
```diff
@@ -24,3 +24,21 @@ if (logoutBttn) {
...
+  document.querySelector('.modal').close()
+
+  // Display error response to user.
+  const error = JSON.parse(e.detail.xhr.response)
+  document.getElementById('toast')
+  .insertAdjacentHTML('beforeend', `
+  <div class="alert alert-error">
+    <i class="fa-solid fa-circle-xmark"></i>
+    <span>${error.error}</span>
```
daniellemaxwell, line 38: Should we review the error responses to ensure users will understand how to resolve issues?

bbengfort: Probably, but for now, having any error that they can contact us at support is helpful.

pkg/web/templates/layouts/base.html
bbengfort: I wasn't having much luck positioning the toasts either. In the counterparty modal, where does the toast appear, in the modal or outside it under main?

bbengfort (approved):
Looks great! I tried to answer your questions but wasn't 100% sure I was on base.

PR #91, opened 2024-05-10
Add accept transaction page

daniellemaxwell:
### Scope of changes

Adds a page that will be displayed after users click accept from the transaction list actions menu.

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/27557431?key=4ef7c77aff19417f6d75254e5a60c233

Note: The link was added to `Archive` in development as seen in the video, but has been moved to `Accept` in the menu.

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/web/routes.go
+++ b/pkg/web/routes.go
@@ -147,6 +147,7 @@ func (s *Server) setupRoutes() (err error) {
 			transactions.GET("/:id", authorize(permiss.TravelRuleView), s.TransactionDetail)
 			transactions.PUT("/:id", authorize(permiss.TravelRuleManage), s.UpdateTransaction)
 			transactions.DELETE("/:id", authorize(permiss.TravelRuleDelete), s.DeleteTransaction)
+			transactions.GET("/:id/accept", authorize(permiss.TravelRuleView), s.TransactionAcceptDetail)
 
 			// Primarily UI methods but are also API Helper Methods
 			transactions.POST("/prepare", authorize(permiss.TravelRuleManage), s.PrepareTransaction)
--- a/pkg/web/routes.go
+++ b/pkg/web/routes.go
@@ -90,6 +90,7 @@ func (s *Server) setupRoutes() (err error) {
 	s.router.GET("/login", s.LoginPage)
 	s.router.GET("/logout", s.Logout)
 	s.router.GET("/transactions", authenticate, s.Transactions)
+	s.router.GET("/transactions/accept/:id", authenticate, s.TransactionsAccept)
 	s.router.GET("/accounts", authenticate, s.Accounts)
 	s.router.GET("/counterparty", authenticate, s.CounterpartyVasps)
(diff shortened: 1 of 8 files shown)


pkg/web/routes.go
```diff
@@ -146,6 +147,7 @@ func (s *Server) setupRoutes() (err error) {
 			transactions.GET("/:id", authorize(permiss.TravelRuleView), s.TransactionDetail)
 			transactions.PUT("/:id", authorize(permiss.TravelRuleManage), s.UpdateTransaction)
 			transactions.DELETE("/:id", authorize(permiss.TravelRuleDelete), s.DeleteTransaction)
+			transactions.GET("/:id/accept", authorize(permiss.TravelRuleView), s.TransactionAcceptDetail)
```
bbengfort, line 150: Is this a preview route or does this actually trigger the accept? If the former, we should change the name to "Preview" if the latter, we should change it as follows:

```suggestion
			transactions.GET("/:id/accept", authorize(permiss.TravelRuleManage), s.AcceptTransaction)
```

pkg/web/static/js/transactionAccept.js
```diff
@@ -0,0 +1,4 @@
+// Get the transaction ID from the URL and add it to the htmx request.
+const id = window.location.pathname.split('/').pop()
+document.getElementById('transaction-accept').setAttribute('hx-get', `/v1/transactions/${id}/accept`)
+
```
bbengfort, line 4: Wouldn't it be better to compute this URL in the view (e.g. using a Go method) rather than with Javascript like this? I think that would be more maintainable in the long run.

pkg/web/routes.go
```diff
@@ -90,6 +90,7 @@ func (s *Server) setupRoutes() (err error) {
 	s.router.GET("/login", s.LoginPage)
 	s.router.GET("/logout", s.Logout)
 	s.router.GET("/transactions", authenticate, s.Transactions)
+	s.router.GET("/transactions-accept/:id", authenticate, s.TransactionsAccept)
```
bbengfort, line 93: In REST parlance, "accept" is an action, so this URL should probably be:

```suggestion
	s.router.GET("/transactions/:id/accept", authenticate, s.TransactionsAcceptPreview)
```

I'm trying to keep things pretty formal since this code has an open source audience. Also we'll probably have the transactions detail page at `/transactions/:id` so this keeps things consistent.

bbengfort (commented):
Couple of questions and suggestions; happy to discuss on Slack if you'd like.

pkg/web/routes.go
daniellemaxwell: This will be the preview.

bbengfort (approved):
:shipit:

PR #95, opened 2024-05-15
Add edit counterparty page

daniellemaxwell:
### Scope of changes

Replaces edit counterparty modal with a page. A bug prevented the list of countries to appear in the edit counterparty modal, as it does in the add counterparty modal. This bug only affected the edit counterparty modal, so creating a separate page felt like the best solution at this time.

Fixes SC-24723

### Type of change

- [x] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/27707197?key=ae327ae4de154133e8dfa15127dd7665

### Author checklist

### Reviewer(s) checklist


pkg/web/static/js/countriesList.js
```diff
@@ -0,0 +1,247 @@
+export const countries = {
```
daniellemaxwell, line 1: This object has been moved to a separate file so that it may be imported in other `js` files.

pkg/web/static/js/editCounterparty.js
```diff
@@ -0,0 +1,19 @@
...
+    settings: {
+      contentLocation: document.getElementById('country-list'),
+    },
+  });
+
+  // Get the selected country value.
+  const cpartyCountry = document.getElementById('selected-country');
+  const cpartySelectedCountry = cpartyCountry.value;
+
+  const countriesArray = Object.entries(countries).map(([value, text]) => ({ text, value }));
```
pdeziel, line 16: Is the countries list referenced by key somewhere else? I'm wondering why we have to create the inverse mapping here every time the DOM loads rather than having that mapping defined in the file.

pdeziel (approved):
Looks good to me

pkg/web/static/js/editCounterparty.js
daniellemaxwell: That's a good observation. I should be able to move `countriesArray` to `countries.js` and then import it here and in other files where it's used.

PR #111, opened 2024-05-27
Return IVMS 101 Payload in Accept Preview

bbengfort:
### Scope of changes

This PR returns the latest secure envelope (decrypted) for the transaction accept preview.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Mostly just check to make sure this will work for the front-end. Do you need the transaction data as well, or just the payload? 

Note: this is going to be tough to test except in production.

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/store/sqlite/transactions.go
+++ b/pkg/store/sqlite/transactions.go
@@ -168,6 +168,39 @@ func (s *Store) DeleteTransaction(ctx context.Context, id uuid.UUID) (err error)
 	return tx.Commit()
 }
 
+const (
+	latestSecEnvSQL            = "SELECT * FROM secure_envelopes WHERE id=:envelopeID ORDER BY timestamp DESC LIMIT 1"
+	latestSecEnvByDirectionSQL = "SELECT * FROM secure_envelopes WHERE id=:envelopeID AND direction=:direction ORDER BY timestamp DESC LIMIT 1"
+)
+
+func (s *Store) LatestSecureEnvelope(ctx context.Context, envelopeID uuid.UUID, direction string) (env *models.SecureEnvelope, err error) {
+	var tx *sql.Tx
+	if tx, err = s.BeginTx(ctx, &sql.TxOptions{ReadOnly: true}); err != nil {
+		return nil, err
+	}
+	defer tx.Rollback()
+
+	var result *sql.Row
+	if direction == "" || direction == models.DirectionAny {
+		// Get the latest secure envelope regardless of the direction
+		result = tx.QueryRow(latestSecEnvSQL, sql.Named("envelopeID", envelopeID))
+	} else {
+		// Specify the direction to get the latest envelope for
+		result = tx.QueryRow(latestSecEnvByDirectionSQL, sql.Named("envelopeID", envelopeID), sql.Named("direction", direction))
+	}
+
+	env = &models.SecureEnvelope{}
+	if err = env.Scan(result); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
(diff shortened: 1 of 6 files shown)


pkg/store/sqlite/transactions.go
```diff
@@ -168,6 +168,39 @@ func (s *Store) DeleteTransaction(ctx context.Context, id uuid.UUID) (err error)
 	return tx.Commit()
 }
 
+const (
+	latestSecEnvSQL            = "SELECT * FROM secure_envelopes WHERE id=:envelopeID ORDER BY timestamp DESC LIMIT 1"
+	latestSecEnvByDirectionSQL = "SELECT * FROM secure_envelopes WHERE id=:envelopeID AND direction=:direction ORDER BY timestamp DESC LIMIT 1"
+)
+
+func (s *Store) LatestSecureEnvelope(ctx context.Context, envelopeID uuid.UUID, direction string) (env *models.SecureEnvelope, err error) {
```
bbengfort, line 176: This code will (hopefully) get the latest secure envelope from the database. The problem is that it's not tested yet.

pkg/web/transactions.go
```diff
@@ -311,16 +314,46 @@ func (s *Server) AcceptTransactionPreview(c *gin.Context) {
...
+
+	// Create the web response, which also includes the transaction details.
+	webData := gin.H{
+		"envelope":    out,
+		"transaction": nil,
+	}
+
+	if webData["transaction"], err = api.NewTransaction(transaction); err != nil {
+		c.Error(err)
+	}
```
bbengfort, line 350: Here's the real question -- the API just gets the secure envelope, but I'm also returning the `*models.Transaction` for the webdata. However, if you don't need this, I'd rather exclude it. The envelope has a `Transaction` field (or a `Pending` field) with the data about the chain, originator and beneficiary crypto addresses, amount, etc. The EnvelopeID is the same as the TransactionID. The information you'd be losing would be `source`, `status`, `counterparty`, `counterparty_id`, and `envelope_count`. Any thoughts?

daniellemaxwell: We have the ability to check the source to determine if `accept` or `reject` should be displayed as an action in the transaction list, so I think that not returning the `*models.Transaction` here will be fine. Other info like `status`, `envelope_count`, etc. is still provided to `transaction_detail.html` so users may view it on the detail page or the list page.

pkg/store/sqlite/transactions.go
```diff
@@ -168,6 +168,39 @@ func (s *Store) DeleteTransaction(ctx context.Context, id uuid.UUID) (err error)
...
+	var tx *sql.Tx
+	if tx, err = s.BeginTx(ctx, &sql.TxOptions{ReadOnly: true}); err != nil {
+		return nil, err
+	}
+	defer tx.Rollback()
+
+	var result *sql.Row
+	if direction == "" || direction == models.DirectionAny {
```
daniellemaxwell, line 184: Pretty sure my brain mixed up direction and source after I responded to your question, but it's all clear again.

daniellemaxwell (approved):
Looks good! I was mixing up transactions and envelopes for a while, essentially thinking they were the same even though we have a `# of envelopes field` in the transaction list. Reviewing this, along with some other changes, has helped solidify the differences for me. Thanks for pushing on all of the Envoy changes!

pkg/web/transactions.go
bbengfort: Ok, in that case I'm going to remove it -- which will definitely simplify things!

PR #137, opened 2024-06-19
Display last update timestamp in local time

daniellemaxwell:
### Scope of changes

This creates a method to localize the `LastUpdate` timestamp as it is currently stored in UTC.

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

The 1st envelope to appear in the table shown in the image was created at 6:05 PM ET.
https://tinyurl.com/2yphxrxz

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/web/templates/partials/transaction/transaction_list.html
+++ b/pkg/web/templates/partials/transaction/transaction_list.html
@@ -109,10 +109,8 @@
         {{ $created := .Created.Format "January 2, 2006 15:04:05" }}
         <td>{{ $created }}</td>
         {{ if .LastUpdate }}
-        {{ $lastUpdate := .LastUpdate.Format "January 2, 2006 15:04:05" }}
+        {{ $lastUpdate := .LocalizeLastUpdate.Format "January 2, 2006 15:04:05" }}
         <td class="trans-last-update">{{ $lastUpdate }}</td>
-        <!-- Place last update in hidden input for debugging. -->
-        <input type="hidden" value="{{ $lastUpdate }}" />
         {{ else }}
         <td>&mdash;</td>
         {{ end }}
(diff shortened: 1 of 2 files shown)


pkg/web/templates/partials/transaction/transaction_list.html
```diff
@@ -109,10 +109,8 @@
         {{ $created := .Created.Format "January 2, 2006 15:04:05" }}
         <td>{{ $created }}</td>
         {{ if .LastUpdate }}
-        {{ $lastUpdate := .LastUpdate.Format "January 2, 2006 15:04:05" }}
+        {{ $lastUpdate := .LocalizeLastUpdate.Format "January 2, 2006 15:04:05" }}
```
daniellemaxwell, line 112: In my timezone, I would see `in 4 hours` in the `last update` column after sending an envelope. Localizing the time should resolve this issue.

bbengfort (commented):
Unfortunately I don't think this method will work unless we can get the timezone information from the user (e.g. maybe we set it as a cookie somehow so that the server knows what the user's timezone is?). 

The central problem is that Javascript is treating the lat update timestamp as the same timezone as the user, so when it does time math, it is incorrect (unless the user is in UTC time). If we added the timezone to the timestamp, would Javascript handle it correctly?

pkg/web/templates/partials/transaction/transaction_list.html
daniellemaxwell: Using the offset appears to work. I took the following screenshot a few minutes after sending a new envelope: https://tinyurl.com/29a7kmh3

bbengfort (approved):
Glad that fixed it!

PR #139, opened 2024-06-20
Amend address range and update transaction accept form data modification

daniellemaxwell:
### Scope of changes

- Removes individual inputs for address (ex. `Street Name` and `City`) and displays data on multiple address lines.
- Modifies request params to ensure addresses are added to the `address_line` array.
- Fixes the identity payload format in the params.
- Checks if `parseFloat` returns a `NaN` and doesn't modify the transaction amount if it does.
- Add margin to form sections.

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/28849328?key=08df42db7543fe4def7a74b6bb0c1867

https://www.awesomescreenshot.com/s/folder/F081jmL5J2/9c2ee19d2ef25dedc8d9e406e5ce6fc9

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/web/static/js/transactionAcceptPreview.js
+++ b/pkg/web/static/js/transactionAcceptPreview.js
@@ -78,7 +78,7 @@ function setIdentifierData(el, identifier) {
 }
 
 // Modify parameters sent in the body of a request via the htmx configRequest.
-const idEl = document.getElementById('send-id')
+const idEl = document.getElementById('envelope-id')
 const id = idEl?.value
 document.body.addEventListener('htmx:configRequest', (e) => {
   const transactionSendEP = `/v1/transactions/${id}/send`;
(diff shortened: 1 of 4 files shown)


pkg/web/templates/transactions_accept.html
```diff
@@ -23,56 +24,4 @@ <h1 class="font-semibold text-2xl md:text-3xl text-center text-balance">Accept T
 {{ define "appcode" }}
 <script src="https://unpkg.com/slim-select@latest/dist/slimselect.min.js"></script>
 <script type="module" src="/static/js/transactionAcceptPreview.js"></script>
-<script>
```
daniellemaxwell, line 26: The script was moved to the `transactionAcceptPreview` file and modified.

pkg/web/static/js/transactionAcceptPreview.js
```diff
@@ -75,4 +75,174 @@ function setIdentifierData(el, identifier) {
...
+          break;
+        // Set the beneficiary VASP country of registration.
+        case key.startsWith('ctry_benf'):
+          beneficiaryVASP[newKey] = params[key];
+          break;
+      };
+    };
+
+    // Convert transaction amount to a float. If conversion fails, set amount to the original value.
+    const amount = parseFloat(data.transaction.amount);
```
pdeziel, line 244: Is the `parseFloat` function unique to this codebase or a JS thing? I'm wondering under what conditions the type conversion would fail, and is that possible to validate before trying to convert it?. Possibly this is just me not understanding the context for which this PR fits into the larger picture.

pdeziel (commented):
Looks good so far, great work parsing through the nested madness that is IVMS. I did have a question below that I'm interested in hearing your opinion on so I didn't mark this as approved yet.

pkg/web/static/js/transactionAcceptPreview.js
daniellemaxwell: `parseFloat` is a JS function that will return the floated value of a string or NaN if the string can't be converted to a number. Here's a [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat) for more more info. 

When submitting the form, the `amount` must be sent as a float or a parse error will be returned from the BE.

pdeziel: Ok, is it possible to return an error to the user when the form is submitted instead of silently failing?

pdeziel: Gotcha, we should try to catch that in the frontend if possible to avoid unnecessary requests to the backend. If we don't have that kind of validation yet on the frontend I understand, but let's create a story for that at least if it's too much for this PR.

pdeziel (approved):

pkg/web/static/js/transactionAcceptPreview.js
daniellemaxwell: I'll create a story so that this may be included when validation gets added to the other forms.

PR #141, opened 2024-06-21
Fix bug duplicating network IDs when wallets are deleted

daniellemaxwell:
### Scope of changes

This PR resolves a bug that caused IDs assigned to additional crypto wallet network elements to duplicate if a wallet was deleted out of sequential order.

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/28830133?key=56e6e5838312f2e6eade6933ceb29ad1

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/web/static/js/customerAccounts.js
+++ b/pkg/web/static/js/customerAccounts.js
@@ -42,8 +42,8 @@ document.body.addEventListener("htmx:configRequest", (e) => {
   }
 });
 
-// Count the number of wallet addresses to ensure a unique number as crypto wallets or added and deleted.
-// Start the counter at 1 to avoid a conflict with the first crypto wallet.
+// Count number of wallet addresses to create a unique ID as crypto wallets are added and deleted.
+// Start counter at 1 to avoid conflict with the first crypto wallet.
 let walletIdCounter = 1;
 
 // Add a new wallet address and network field to the new customer account form modal on click.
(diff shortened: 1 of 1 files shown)


pkg/web/static/js/customerAccounts.js
```diff
@@ -42,9 +42,13 @@ document.body.addEventListener("htmx:configRequest", (e) => {
   }
 });
 
+// Count number of wallet addresses to create a unique ID as crypto wallets are added and deleted.
+// Start counter at 1 to avoid conflict with the first crypto wallet.
+let walletIdCounter = 1;
```
pdeziel, line 47: If we need unique IDs, can we use an ID generation library here? Such as https://www.npmjs.com/package/uuid. That would avoid having to rely on the global variable.

pdeziel: Alternatively you might be able to use https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID and avoid the dependency.

pdeziel (commented):
Looks good to me so far, I had a suggestion below about generating IDs that might be easier than using the global counter.

pkg/web/static/js/customerAccounts.js
daniellemaxwell: I think that's a much better idea. I'll make the update.

pdeziel (approved):
Thanks for making the changes! Looks a lot simpler!

PR #150, opened 2024-07-04
Change User Password

bbengfort:
### Scope of changes

Adds an endpoint that allows users to change the password of other users.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Allows a password change to occur.

I think that we should create a story to modify this endpoint to email the user the new password once we have an email configuration setup.

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/cmd/envoy/main.go
+++ b/cmd/envoy/main.go
@@ -20,7 +20,7 @@ import (
 	"github.com/trisacrypto/envoy/pkg/store/models"
 	"github.com/trisacrypto/envoy/pkg/store/sqlite"
 	"github.com/trisacrypto/envoy/pkg/ulids"
-	"github.com/trisacrypto/envoy/pkg/web/auth"
+	"github.com/trisacrypto/envoy/pkg/web/auth/passwords"
 	permiss "github.com/trisacrypto/envoy/pkg/web/auth/permissions"
 
 	"github.com/trisacrypto/envoy/pkg/config"
@@ -422,8 +422,8 @@ func createUser(c *cli.Context) (err error) {
 		return cli.Exit("specify admin, compliance, or observer as the role", 1)
 	}
 
-	password := auth.AlphaNumeric(12)
-	if user.Password, err = auth.CreateDerivedKey(password); err != nil {
+	password := passwords.AlphaNumeric(12)
+	if user.Password, err = passwords.CreateDerivedKey(password); err != nil {
 		return cli.Exit(err, 1)
 	}
 
@@ -447,8 +447,8 @@ func resetPassword(c *cli.Context) (err error) {
 		return cli.Exit(fmt.Errorf("could not retrieve user %q: %w", c.String("email"), err), 1)
 	}
 
-	password := auth.AlphaNumeric(12)
-	if user.Password, err = auth.CreateDerivedKey(password); err != nil {
+	password := passwords.AlphaNumeric(12)
+	if user.Password, err = passwords.CreateDerivedKey(password); err != nil {
 		return cli.Exit(fmt.Errorf("could not create derived key for user password: %w", err), 1)
 	}
 
@@ -466,11 +466,11 @@ func createAPIKey(c *cli.Context) (err error) {
 	}
 
 	key := &models.APIKey{
-		ClientID: auth.KeyID(),
+		ClientID: passwords.KeyID(),
 	}
 
-	secret := auth.Secret()
-	if key.Secret, err = auth.CreateDerivedKey(secret); err != nil {
+	secret := passwords.Secret()
+	if key.Secret, err = passwords.CreateDerivedKey(secret); err != nil {
 		return cli.Exit(err, 1)
 	}
 
--- a/cmd/envoy/main.go
+++ b/cmd/envoy/main.go
@@ -452,8 +452,8 @@ func resetPassword(c *cli.Context) (err error) {
 		return cli.Exit(fmt.Errorf("could not create derived key for user password: %w", err), 1)
 	}
 
-	if err = db.UpdateUser(ctx, user); err != nil {
-		return cli.Exit(fmt.Errorf("could not update user %s: %w", user.Email, err), 1)
+	if err = db.SetUserPassword(ctx, user.ID, user.Password); err != nil {
+		return cli.Exit(fmt.Errorf("could not store password: %w", err), 1)
 	}
 
 	fmt.Printf("updated user %s with password: %s\n", user.Email, password)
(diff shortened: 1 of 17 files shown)


cmd/envoy/main.go
```diff
@@ -452,8 +452,8 @@ func resetPassword(c *cli.Context) (err error) {
 		return cli.Exit(fmt.Errorf("could not create derived key for user password: %w", err), 1)
 	}
 
-	if err = db.UpdateUser(ctx, user); err != nil {
-		return cli.Exit(fmt.Errorf("could not update user %s: %w", user.Email, err), 1)
+	if err = db.SetUserPassword(ctx, user.ID, user.Password); err != nil {
```
bbengfort, line 455: This was a bug … that I'm surprised I haven't heard about yet …

pkg/web/api/v1/user.go
```diff
@@ -24,6 +25,11 @@ type UserList struct {
 	Users []*User    `json:"users"`
 }
 
+type UserPassword struct {
+	Password  string `json:"password"`
+	SendEmail bool   `json:"send_email"`
```
bbengfort, line 30: `SendEmail` currently doesn't do anything, but in the future I'm thinking that if this is true, Envoy could send the user an email with a password change notice.

pkg/web/api/v1/user.go
```diff
@@ -104,3 +109,46 @@ func (u *User) Model() (model *models.User, err error) {
 
 	return model, nil
 }
+
+func (u UserPassword) Validate() (err error) {
```
bbengfort, line 113: Password strength validation is here … not sure if this is the right place for it or not.

pkg/web/auth.go
```diff
@@ -341,3 +342,76 @@ func (s *Server) reauthenticateAPIKey(c *gin.Context, keyID ulid.ULID) (_ *auth.
 
 	return auth.NewClaims(ctx, apikey)
 }
+
+func (s *Server) ChangePassword(c *gin.Context) {
```
bbengfort, line 346: This handler allows the logged in user to change their own password, regardless of their role.

pkg/web/routes.go
```diff
@@ -125,6 +125,9 @@ func (s *Server) setupRoutes() (err error) {
 		v1.POST("/authenticate", s.Authenticate)
 		v1.POST("/reauthenticate", s.Reauthenticate)
 
+		// User Profile Management
+		v1.POST("/change-password", authenticate, s.ChangePassword)
```
bbengfort, line 129: This route allows the logged in user to change their password -- I'm thinking it is part of a suite of user profile tools?

pkg/web/routes.go
```diff
@@ -195,6 +198,7 @@ func (s *Server) setupRoutes() (err error) {
 			users.GET("/:id", authorize(permiss.UsersView), s.UserDetail)
 			users.PUT("/:id", authorize(permiss.UsersManage), s.UpdateUser)
 			users.DELETE("/:id", authorize(permiss.UsersManage), s.DeleteUser)
+			users.POST("/:id/password", authorize(permiss.UsersManage), s.ChangeUserPassword)
```
bbengfort, line 201: This is the API-based user management change password. Any user's password can be changed by anyone with the UsersManage permission (Admin role)

pkg/web/users.go
```diff
@@ -290,8 +291,66 @@ func (s *Server) DeleteUser(c *gin.Context) {
...
 	c.Negotiate(http.StatusOK, gin.Negotiate{
 		Offered:  []string{binding.MIMEJSON, binding.MIMEHTML},
-		HTMLData: gin.H{"UserID": userID},
+		HTMLData: scene.Scene{"UserID": userID},
 		JSONData: api.Reply{Success: true},
 		HTMLName: "user_delete.html",
 	})
 }
+
+func (s *Server) ChangeUserPassword(c *gin.Context) {
```
bbengfort, line 300: This handler is the Users API resource action to change a specific users's password.

pkg/web/accounts.go
```diff
@@ -281,7 +282,7 @@ func (s *Server) DeleteAccount(c *gin.Context) {
 	// Content negotiation
 	c.Negotiate(http.StatusOK, gin.Negotiate{
 		Offered:  []string{binding.MIMEJSON, binding.MIMEHTML},
-		HTMLData: gin.H{"AccountID": accountID},
+		HTMLData: scene.Scene{"AccountID": accountID},
```
daniellemaxwell, line 285: I should've asked when I didn't see `scene.Scene` used in the users PR. I thought it meant we shouldn't update it where `HTMLData` was used. Thank you for making the changes.

pkg/web/api/v1/user.go
daniellemaxwell: Would it be better to put it in `auth`?

pkg/web/routes.go
daniellemaxwell: This would be used on a user profile page (that's specific to the logged in user) while the API route below would be used on a users management page that would make the option to change passwords (and other settings in the future) only available to the admin?

daniellemaxwell (approved):
Nicely done! I appreciated the comments providing more detail on your thought process. I have a couple of questions, if you could take a look. This is approved and may be merged when you're ready.

pkg/web/api/v1/user.go
bbengfort: Yes, it would be better to put it into `auth`. I'll do that!

pkg/web/routes.go
bbengfort: That's correct!

pkg/web/accounts.go
bbengfort: I don't think scene wasn't merged yet which is why it wasn't in that PR; but we're all caught up now (hopefully).

pkg/web/auth.go
```diff
@@ -341,3 +342,76 @@ func (s *Server) reauthenticateAPIKey(c *gin.Context, keyID ulid.ULID) (_ *auth.
...
+	// Set the password for the specified user
+	if err = s.store.SetUserPassword(c.Request.Context(), userID, derivedKey); err != nil {
+		c.Error(err)
+		c.JSON(http.StatusInternalServerError, api.Error("could not complete change user password request"))
+		return
+	}
+
+	// Create template scene for rendering information about the user
+	data := scene.New(c)
+	data["Success"] = true
```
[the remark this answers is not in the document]
bbengfort: `IsAuthenticated` is more about the logged in user based on the cookie value -- this is just referring to the fact that the operation was successful. This is a JSON API pattern we've adapted for a while, but I imagine that you'll remove this to add a message and a class for a toast -- e.g. "Password for benjamin@rotational.io successfully changed" or something like that.

PR #152, opened 2024-07-05
Add user management page and create user modal 

daniellemaxwell:
### Scope of changes

- Adds route to the user management page
- Updates menu to display user settings
- Displays user list table on user management page
- Admin users will have the ability to add a user
- Users will be able to copy the created password after a new user is successfully added

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/29290658?key=073ded1ec8b48ab35ecdc4e3cf0efc96

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/web/api/v1/user.go
+++ b/pkg/web/api/v1/user.go
@@ -12,9 +12,9 @@ type User struct {
 	ID        ulid.ULID  `json:"id,omitempty"`
 	Name      string     `json:"name"`
 	Email     string     `json:"email"`
-	Passsword string     `json:"password,omitempty"`
+	Password  string     `json:"password,omitempty"`
 	Role      string     `json:"role"`
-	LastLogin *time.Time `json:"last_login"`
+	LastLogin *time.Time `json:"last_login,omitempty"`
 	Created   time.Time  `json:"created,omitempty"`
 	Modified  time.Time  `json:"modified,omitempty"`
 }
@@ -66,7 +66,7 @@ func (u *User) Validate() (err error) {
 		err = ValidationError(err, MissingField("email"))
 	}
 
-	if u.Passsword != "" {
+	if u.Password != "" {
 		err = ValidationError(err, ReadOnlyField("password"))
 	}
 
--- a/pkg/web/pages.go
+++ b/pkg/web/pages.go
@@ -79,3 +79,7 @@ func (s *Server) SendEnvelopeForm(c *gin.Context) {
 func (s *Server) TravelAddressUtility(c *gin.Context) {
 	c.HTML(http.StatusOK, "traveladdress.html", scene.New(c))
 }
+
+func (s *Server) UserSettings(c *gin.Context) {
+	c.HTML(http.StatusOK, "user_settings.html", scene.New(c))
+}
--- a/pkg/web/routes.go
+++ b/pkg/web/routes.go
@@ -103,6 +103,7 @@ func (s *Server) setupRoutes() (err error) {
 	s.router.GET("/counterparty", authenticate, s.CounterpartyVasps)
 	s.router.GET("/send-envelope", authenticate, s.SendEnvelopeForm)
 	s.router.GET("/utilities/travel-address", authenticate, s.TravelAddressUtility)
+	s.router.GET("/user-settings", authenticate, s.UserSettings)
 
 	// Swagger documentation with Swagger UI hosted from a CDN
 	s.router.GET("/v1/docs", gin.WrapH(v5cdn.New(
--- a/pkg/web/templates/components/nav.html
+++ b/pkg/web/templates/components/nav.html
@@ -27,9 +27,10 @@
    <div class="dropdown dropdown-bottom dropdown-left">
     <div tabindex="0" role="button" class="mr-4">
       <i class="fa-solid fa-user text-lg"></i>
-      <span class="sr-only">Open User Settings Menu</span>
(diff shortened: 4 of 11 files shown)


pkg/web/api/v1/user.go
```diff
@@ -13,9 +13,9 @@ type User struct {
 	ID        ulid.ULID  `json:"id,omitempty"`
 	Name      string     `json:"name"`
 	Email     string     `json:"email"`
-	Passsword string     `json:"password,omitempty"`
+	Password  string     `json:"password,omitempty"`
```
daniellemaxwell, line 16: Such an easy to miss typo.

pkg/web/api/v1/user.go
```diff
@@ -13,9 +13,9 @@ type User struct {
...
 	Role      string     `json:"role"`
-	LastLogin *time.Time `json:"last_login"`
+	LastLogin *time.Time `json:"last_login,omitempty"`
```
daniellemaxwell, line 18: I added `omitempty` here after noticing it wasn't included, but it can be removed.

pkg/web/pages.go
```diff
@@ -79,3 +79,7 @@ func (s *Server) SendEnvelopeForm(c *gin.Context) {
 func (s *Server) TravelAddressUtility(c *gin.Context) {
 	c.HTML(http.StatusOK, "traveladdress.html", scene.New(c))
 }
+
+func (s *Server) UserSettings(c *gin.Context) {
```
daniellemaxwell, line 83: I'm not thinking this should be UserManagement since there will be an option for individual users to manage their password and potentially other settings in the future.

pkg/web/templates/partials/user/user_list.html
```diff
@@ -0,0 +1,47 @@
...
+    <tbody>
+      {{ if .Users }}
+      {{ range .Users }}
+      <tr class="hover">
+        <td>{{ .Name }}</td>
+        <td>{{ .Email }}</td>
+        <td>{{ .Role }}</td>
+        {{ $created := .Created.Format "January 2, 2006 15:04:05" }}
+        <td>{{ $created }}</td>
+        {{ $lastLogin := .LastLogin.Format "January 2, 2006 15:04:05" }}
```
daniellemaxwell, line 23: The following error is sometimes returned when refreshing the users list: `error calling Format: value method time.Time.Format called using nil *Time pointer`

I believe there was a similar issue with a previous resource, but I couldn't quite remember how it was resolved.

pkg/web/api/v1/user.go
bbengfort: This is good - I think it should be `omitempty` here.

pkg/web/pages.go
bbengfort: Perhaps we could call this `UserManagement` and call the option for individual users to manage their password `ProfileManagement`?

pkg/web/routes.go
```diff
@@ -103,6 +103,7 @@ func (s *Server) setupRoutes() (err error) {
 	s.router.GET("/counterparty", authenticate, s.CounterpartyVasps)
 	s.router.GET("/send-envelope", authenticate, s.SendEnvelopeForm)
 	s.router.GET("/utilities/travel-address", authenticate, s.TravelAddressUtility)
+	s.router.GET("/user-settings", authenticate, s.UserSettings)
```
bbengfort, line 106: To match the case of the rest of the routes, I recommend it be `/users` -- although by this logic it should also be `/counterparties` above, but we can skip that one. 

I'll make that change if that's ok.

pkg/web/templates/components/nav.html
```diff
@@ -27,9 +27,10 @@
    <div class="dropdown dropdown-bottom dropdown-left">
     <div tabindex="0" role="button" class="mr-4">
       <i class="fa-solid fa-user text-lg"></i>
-      <span class="sr-only">Open User Settings Menu</span>
+      <span class="sr-only">Open Settings Menu</span>
```
bbengfort, line 30: Should we add screen reader instructions for the other menu items? Is that what `sr-only` means? What is the role of the icon above?

pkg/web/templates/partials/user/user_list.html
bbengfort: Yes, you'll have to do the if last login check on the timestamp which can be nil; not on the returned value of the Format call, e.g. 

```suggestion
        {{ if .LastLogin  }}
        {{ .LastLogin.Format "January 2, 2006 15:04:05" }}
        {{ else }}
        <td>&mdash;</td>
        {{ end }}
```

I'll make that change in the PR.

bbengfort (approved):
@daniellemaxwell this looks really great - you did an ace job getting this together quickly and looking fantastic! I appreciate the care you took with the create user form and pre-validation on the front-end. 

I've made some changes based on my review comments in order to get this merged ASAP. If you'd like to discuss any of the changes let me know; we can always create a story to go a different direction if you want.

pkg/web/templates/components/nav.html
daniellemaxwell: Yes that's what `sr-only` stands for. This was added because we only have an icon on the page, so adding this would help explain what the icon does. 

The menu items have text next to the icons that explain each menu option, so that's why `sr-only` wasn't used there.

PR #153, opened 2024-07-05
Change user password

daniellemaxwell:
### Scope of changes

This PR adds a user profile page that allows individual users to reset their password.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/29295116?key=4da41c0530399052d8733ae55f5187a0

### Author checklist

### Reviewer(s) checklist


pkg/web/templates/partials/password/password_changed.html
```diff
@@ -0,0 +1 @@
+<div></div>
```
daniellemaxwell, line 1: A swap didn't seem necessary here, but this file had to be included to prevent a panic.

bbengfort (approved):
Very nice! I think our users will find this helpful when managing passwords across multiple nodes. 

So during the merge to main, I ran into a few bumps, so I had to fix those.

Also per discussion in the previous PR, I renamed this link to just `/profile`. 

I've also added a "confirm password" field and the submit button is only enabled when both passwords match, just to make sure the user doesn't typo their password and then can't log back in!

pkg/web/pages.go
```diff
@@ -80,6 +80,6 @@ func (s *Server) TravelAddressUtility(c *gin.Context) {
 	c.HTML(http.StatusOK, "traveladdress.html", scene.New(c))
 }
 
-func (s *Server) UserProfile(c *gin.Context) {
-	c.HTML(http.StatusOK, "user_profile.html", scene.New(c))
+func (s *Server) UsersManagement(c *gin.Context) {
+	c.HTML(http.StatusOK, "users_management.html", scene.New(c))
```
bbengfort, line 84: lol, git guessed wrong on the merge this time!

pkg/web/static/js/userProfile.js
```diff
@@ -0,0 +1,53 @@
...
+    changePasswordBtn.setAttribute('disabled', 'true');
+  }
+}
+
+// TODO: is this a complete list of events that should be used to detect if the passwords match?
+addEventListeners(passwordInput, 'change keyup paste cut', checkPasswords);
+addEventListeners(confirmPassword, 'change keyup paste cut', checkPasswords);
+
+// Write code to run after an htmx request event.
+// TODO: should we do a pre-flight check as a last sanity check that the passwords match?
```
bbengfort, line 34: @daniellemaxwell I just wanted to call your attention to these TODOs I put in during my review -- let me know what you think about these questions.

pkg/web/static/js/userProfile.js
```diff
@@ -0,0 +1,53 @@
...
+
+// Reset the form when change password modal is closed in case user entered some values without submitting.
+closePwdModalBtn.addEventListener('click', () => {
+  changePwdForm.reset();
+});
+
+// Enable the change password button when the password and confirm password inputs are identical.
+// TODO: should we add any other visual indication that the passwords do not match?
+const checkPasswords = e => {
+  if (passwordInput.value == confirmPassword.value && passwordInput.value != '') {
```
bbengfort, line 22: Should this be `===` and `!==`? I can never remember with Javascript.

daniellemaxwell: `===` and `!==` or `strict equality` should be used. It will ensure the types also match.

pkg/web/static/js/userProfile.js
daniellemaxwell: `setCustomValidity` could be used to provide specific wording to users here: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setCustomValidity

I noticed that the button doesn't return to the disabled state if a user enters a password that doesn't meet the validation requirements. I'll make an update.

pkg/web/static/js/userProfile.js
bbengfort: ok, so only use `==` and `!=` when it can be undefined or `null`?

pkg/web/static/js/userProfile.js
bbengfort: oh, setCustomValidity is awesome - so much better than putting class state on the inputs!

pkg/web/static/js/userProfile.js
daniellemaxwell: Yes. Even then, you'll still often see `===` or `!==` used.

pkg/web/static/js/userProfile.js
daniellemaxwell: I was going to use `setCustomValidity` but the message didn't appear when I clicked the submit button. I must've missed something, so I'll create a story to revisit and make the update.

bbengfort: Cheers!

PR #154, opened 2024-07-08
Allow admin users to reset passwords

daniellemaxwell:
### Scope of changes

This PR allows an admin user to reset the password of a user who has forgotten their password.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/29358530?key=03b7d2d472241dc629128739d839fb94

### Author checklist

### Reviewer(s) checklist


pkg/web/api/v1/user.go
```diff
@@ -30,6 +36,10 @@ type UserPassword struct {
 	SendEmail bool   `json:"send_email"`
 }
 
+type UserQuery struct {
```
daniellemaxwell, line 39: The user query was added so that the FE may make a GET request to display the change password modal. We needed the ability to swap the modal into the `Users Management` page while also being able to access the ID of the user whose password will be changed. Instead of creating a separate route, the query was added similar to what was done in the `TransactionDetail` handler.

pkg/web/api/v1/user.go
```diff
@@ -123,3 +133,20 @@ func (u UserPassword) Validate() (err error) {
...
 }
+
+//===========================================================================
+// User Query
+//===========================================================================
+
+func (q *UserQuery) Validate() (err error) {
+	q.Detail = strings.ToLower(strings.TrimSpace(q.Detail))
+	if q.Detail == "" {
+		q.Detail = DetailUser
```
daniellemaxwell, line 144: Although we aren't currently displaying a user detail modal, it made sense to make that the default going forward.

pkg/web/static/js/userProfile.js
```diff
@@ -1,6 +1,8 @@
+import { setSuccessToast } from "./utils.js";
+
 // Helper function to add the same event handler for multiple events
-const addEventListeners = (el, evts, fn) => {
-  evts.split(' ').forEach(e => el.addEventListener(e, fn, false));
+export const addEventListeners = (el, evts, fn) => {
+  evts.split(' ').forEach(e => el?.addEventListener(e, fn, false));
```
daniellemaxwell, line 5: Optional chaining was added throughout this file to resolve some undefined errors that were appeared in the console.

pkg/web/static/js/usersManagement.js
```diff
@@ -29,6 +33,29 @@ document.addEventListener('htmx:afterSettle', (e) => {
...
+  const userID = document.getElementById('user-id');
+  const userDetailEP = `/v1/users/${userID?.value}?detail=password`;
+
+  if (e.detail.requestConfig.path === userDetailEP && e.detail.requestConfig.verb === 'get') {
+    const passwordInput = document.getElementById('password');
+    const confirmPassword = document.getElementById('confirm-password');
+    const changePasswordBtn = document.getElementById('change-password-btn');
+
+    // TODO: Convert to a reusable function for password matching.
+    const checkPasswords = () => {
```
daniellemaxwell, line 47: I tried to import this from `userProfile` but I think because the DOM changes, the IDs weren't recognized even though I imported those as well. I'll refactor this so that the function doesn't have to be copied.

pkg/web/static/js/utils.js
```diff
@@ -0,0 +1,11 @@
+export function setSuccessToast(msg) {
```
daniellemaxwell, line 1: A follow on story will be created to use this function in other `js` files where a success toast message should be displayed.

pkg/web/templates/user_profile.html
```diff
@@ -57,7 +57,7 @@ <h1 class="font-bold text-xl">Change Password</h1>
           </div>
           <div class="mb-4">
             <label for="confirm-password" class="label-style pb-0.5">Confirm Password</label>
-            <input type="password" id="confirm-password" name="confirm-password" class="input-style" />
```
daniellemaxwell, line 60: This was removed so that `confirm-password` no longer appears as part of the request body. However, if we want to update the BE this can be reversed.

pkg/web/static/js/usersManagement.js
bbengfort: Sounds good - I'm assuming that if the element is looked up after the DOM refresh, it should find it; which means the element needs to be fetched inside the function call?

pkg/web/templates/components/nav.html
```diff
@@ -32,7 +32,9 @@
     <ul tabindex="0" class="dropdown-content z-[1] bg-black text-white shadow menu menu-sm p-2 rounded-box w-52 font-semibold">
       <li><a href="/profile"><i class="fa-solid fa-user-gear"></i>User Profile</a></li>
       <li><a href="/about"><i class="fa-solid fa-gear"></i>Node Configuration</a></li>
+      {{ if .HasRole "Admin" }}
       <li><a href="/users"><i class="fa-solid fa-users"></i>Users Management</a></li>
+      {{ end }}
```
bbengfort, line 37: Let's ask Edwin if we need to guard the whole user management page -- e.g. if it's fine if compliance and observers see the list of users on the node. 

Note that this will not prevent the user from going to `/users` directly, so if the answer to the above question is "no" then we should add an `authorize` to this route that allows only admins to access the page. If the answer is yes, then no backend change is needed.

pkg/web/templates/user_profile.html
bbengfort: that's fine - I didn't really mind it going with the request since the BE just ignores it, but it's best practice to minimize the amount of data transmitted, so thank you for removing it!

pkg/web/users.go
```diff
@@ -172,10 +182,22 @@ func (s *Server) UserDetail(c *gin.Context) {
 		return
 	}
 
+	// Determine the HTML template to render based on the query
```
bbengfort, line 185:
```suggestion
	// Determine the HTML template to render based on the query
	// NOTE: if no query is provided, the UserQuery defaults to 'user'
```

pkg/web/templates/components/nav.html
bbengfort: Is it possible to put this guard around the actions column in the users_management.html page? I realize probably why you opted to put this here was so you didn't have to change every table row with the clickable change password action, which probably has some javascript implications for the modal?

bbengfort (approved):
Looks great, thanks for adding this! I had a couple of comments and suggestions so I'm not going to merge, but feel free to review and merge when ready!

pkg/web/static/js/usersManagement.js
daniellemaxwell: Yes, in this case the element has to be fetched inside an `htmx:afterSettle` event or we'll get an `undefined` response.

pkg/web/templates/components/nav.html
daniellemaxwell: It should be possible. I'll make the change.

pkg/web/templates/user_profile.html
daniellemaxwell: I've applied the suggested change in this [commit](https://github.com/trisacrypto/envoy/pull/154/commits/a8c8ac5619acaf6b5904708376bc306d3bc5bf38). I'll go ahead and merge, but if you think it should be handled differently let me know and I can make an update.

I should note that I tried to return a bool instead of a string, but got an error preventing the condition from running on line 30 of the `user_list.html` partial.

pkg/web/templates/components/nav.html
daniellemaxwell: I should note that I tried return a bool instead of a string, but got an error preventing the condition from running on line 30 of the user_list.html` partial.

PR #156, opened 2024-07-10
Displays account and counterparty edit actions to users with appropriate role

daniellemaxwell:
### Scope of changes

This PR amends the account and counterparty list actions so that the option to edit only appears to users who have the ability to complete the action. The `UserList` has been amended to return a bool instead of a string to check that a user has a specified role.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/29428358?key=7fd0c063b163919aa7c763572886b6f5

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/web/api/v1/account.go
+++ b/pkg/web/api/v1/account.go
@@ -42,9 +42,8 @@ type CryptoAddress struct {
 }
 
 type AccountsList struct {
-	Page          *PageQuery `json:"page"`
-	Accounts      []*Account `json:"accounts"`
-	HasPermission bool       `json:"permission"`
+	Page     *PageQuery `json:"page"`
+	Accounts []*Account `json:"accounts"`
 }
 
 type CryptoAddressList struct {
--- a/pkg/web/api/v1/account.go
+++ b/pkg/web/api/v1/account.go
@@ -42,8 +42,9 @@ type CryptoAddress struct {
 }
 
 type AccountsList struct {
-	Page     *PageQuery `json:"page"`
-	Accounts []*Account `json:"accounts"`
+	Page          *PageQuery `json:"page"`
+	Accounts      []*Account `json:"accounts"`
+	HasPermission bool       `json:"permission"`
 }
 
 type CryptoAddressList struct {
--- a/pkg/web/api/v1/user.go
+++ b/pkg/web/api/v1/user.go
@@ -27,9 +27,9 @@ type User struct {
 }
 
 type UserList struct {
-	Page         *PageQuery `json:"page"`
-	Users        []*User    `json:"users"`
-	AuthUserRole string     `json:"user_role"`
+	Page    *PageQuery `json:"page"`
+	Users   []*User    `json:"users"`
+	IsAdmin bool       `json:"is_admin"`
 }
 
 type UserPassword struct {
--- a/pkg/web/api/v1/user.go
+++ b/pkg/web/api/v1/user.go
@@ -27,9 +27,8 @@ type User struct {
 }
 
 type UserList struct {
-	Page    *PageQuery `json:"page"`
-	Users   []*User    `json:"users"`
-	IsAdmin bool       `json:"is_admin"`
+	Page  *PageQuery `json:"page"`
+	Users []*User    `json:"users"`
 }
 
 type UserPassword struct {
--- a/pkg/web/scene/scene.go
+++ b/pkg/web/scene/scene.go
@@ -20,6 +20,9 @@ const (
 	Page            = "Page"
 	IsAuthenticated = "IsAuthenticated"
 	User            = "User"
+	UserAdmin       = "Admin"
+	UserCompliance  = "Compliance"
+	UserObserver    = "Observer"
 )
 
 type Scene map[string]interface{}
@@ -84,4 +87,9 @@ func GetAuthUserRole(c *gin.Context) string {
 	ctx := New(c)
 	user := ctx.GetUser()
 	return user.Role
+}
+
+func HasPermission(c *gin.Context) bool {
+	role := GetAuthUserRole(c)
+	return role != UserObserver
 }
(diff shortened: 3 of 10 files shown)


pkg/web/api/v1/account.go
```diff
@@ -42,8 +42,9 @@ type CryptoAddress struct {
 }
 
 type AccountsList struct {
-	Page     *PageQuery `json:"page"`
-	Accounts []*Account `json:"accounts"`
+	Page          *PageQuery `json:"page"`
+	Accounts      []*Account `json:"accounts"`
+	HasPermission bool       `json:"permission"`
```
daniellemaxwell, line 47: Amending the response seemed like the best way to check a user's role or if they have permission to perform certain actions in the partial files.

bbengfort: I do get the motivation to do this - particularly as the route handler is the thing that does the permission checking. The problem is that amending this response will also return that bool to API users and permissions in API responses isn't best practice or at the very least would require extensive documentation. 

I think we need to delegate all permissions operations to the `scene`, and that might mean we have to update all of our content negotiation as follows:

```golang
	c.Negotiate(http.StatusOK, gin.Negotiate{
		Offered:  []string{binding.MIMEJSON, binding.MIMEHTML},
		Data:     out,
		HTMLData: scene.New().WithContext(out).WithPermission(foo),
		HTMLName: "account_list.html",
	})
```

Where `WithContext` allows the scene to capture the API context and then we add our permission handling to the scene instead of the api response. What do you think about this approach?

pkg/web/scene/scene.go
```diff
@@ -85,3 +93,8 @@ func GetAuthUserRole(c *gin.Context) string {
 	user := ctx.GetUser()
 	return user.Role
 }
+
+func HasPermission(c *gin.Context) bool {
+	role := GetAuthUserRole(c)
+	return role != UserObserver
+}
```
bbengfort, line 100: Very clever!

pkg/web/api/v1/account.go
daniellemaxwell: Yeah, that was my concern. When I tried sending the context via `HTMLData` none of the data appeared and I wasn't sure if we wanted to do it this way or add the data to the context, though I didn't think of changing the content negotiation in this way. I'm fine with adopting this approach.

pkg/web/api/v1/user.go
```diff
@@ -27,9 +27,8 @@ type User struct {
 }
 
 type UserList struct {
-	Page         *PageQuery `json:"page"`
-	Users        []*User    `json:"users"`
-	AuthUserRole string     `json:"user_role"`
```
bbengfort, line 32: What was `AuthUserRole`?

pkg/web/scene/scene.go
```diff
@@ -49,15 +51,32 @@ func New(c *gin.Context) Scene {
...
+	return s
 }
 
-func (s Scene) HasRole(role string) bool {
-	if user := s.GetUser(); user != nil {
-		return user.Role == role
+func (s Scene) WithAPIData(data interface{}) Scene {
+	s[APIData] = data
+	return s
+}
```
bbengfort, line 64: This is the new chaining function that wraps the api data.

pkg/web/scene/scene.go
```diff
@@ -73,15 +92,48 @@ func (s Scene) GetUser() *auth.Claims {
...
-	ctx := New(c)
-	user := ctx.GetUser()
-	return user.Role
+func (s Scene) IsAdmin() bool {
+	return s.HasRole(RoleAdmin)
+}
+
+func (s Scene) IsViewOnly() bool {
+	return s.HasRole(RoleObserver)
+}
```
bbengfort, line 108: Tried to simplify your original permissions functions a bit on the Scene.

pkg/web/scene/scene.go
```diff
@@ -73,15 +92,48 @@ func (s Scene) GetUser() *auth.Claims {
...
+	return nil
+}
+
+func (s Scene) CounterpartyList() *api.CounterpartyList {
+	if data, ok := s[APIData]; ok {
+		if out, ok := data.(*api.CounterpartyList); ok {
+			return out
+		}
+	}
+	return nil
```
bbengfort, line 138: Unfortunately these are now required in order to get the api data out of scene into the template using a `{{ with }}` block.

pkg/web/templates/partials/account/account_list.html
```diff
@@ -1,3 +1,5 @@
+{{ $canEditAccounts := not .IsViewOnly }}
+{{- with .AccountsList -}}
```
bbengfort, line 2: Here is how to extract the API data from the scene. WE'll have to write the type assertion function for every APIData type we add.

pkg/web/templates/partials/user/user_list.html
```diff
@@ -1,4 +1,5 @@
-{{ $role := .AuthUserRole }}
+{{- $isAdmin := .IsAdmin -}}
```
bbengfort, line 1: This has to be moved to the top since the `with` overwrites the local context with the user list.

pkg/web/api/v1/user.go
daniellemaxwell: I previously returned a string instead of a bool when I initially updated the UserList.

pkg/web/templates/partials/account/account_list.html
daniellemaxwell: I definitely wouldn't have known I'd need to include `-` inside the brackets on line 2.

bbengfort: You don't have to technically - those are whitespace eliminators; so if you put `{{-` that will eat all the whitespace that comes before the `{{` and if you add `-}}` it will eat all the whitespace after the tag. If you didn't include them, the whitespace would remain there, which isn't really a big deal in HTML.

daniellemaxwell: Ah, ok.

PR #168, opened 2024-07-19
Javascript time localization

bbengfort:
### Scope of changes

Uses a Javascript function to localize datetime inside of Envoy.

https://www.awesomescreenshot.com/video/29721214?key=1d6505650ad5b2ce6a2318b9ebc5041e

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [x] other (localization)

### Acceptance criteria

1. Do the event listeners make sense? Should we only do it on `htmx:afterSwap`? Should we put the function call on the partial instead so it's called when loaded?
2. Does the formatting meet with our design expectations?
3. Are all datetimes handled correctly or are there others?

### Author checklist

### Reviewer(s) checklist

- [ ] The event listeners are appropriate/timing of the datetime conversion makes sense
- [ ] The formatting options to Intl.DateTimeFormat look good

### DIFF
--- a/pkg/web/static/js/main.js
+++ b/pkg/web/static/js/main.js
@@ -48,8 +48,7 @@ document.body.addEventListener('htmx:responseError', (e) => {
 /*
  * Timestamp handling: localizes and formats datetimes on the page.
  */
-const updateDatetimes = (e) => {
-  const options = {};
+const updateDatetimes = () => {
   const dtfmt = Intl.DateTimeFormat(navigator.languages, {
     year: 'numeric', month: 'long', day: 'numeric',
     hour: 'numeric', minute:'numeric', second: 'numeric',
--- a/pkg/web/static/js/main.js
+++ b/pkg/web/static/js/main.js
@@ -58,9 +58,11 @@ const updateDatetimes = (e) => {
 
   const datetimes = document.querySelectorAll('.datetime');
   datetimes?.forEach(elem => {
-    const dt = new Date(elem.textContent);
-    elem.textContent = dtfmt.format(dt);
-    elem.classList.remove('datetime');
+    if (elem.textContent !== '' || elem.textContent !== null) {
+      const dt = new Date(elem.textContent);
+      elem.textContent = dtfmt.format(dt);
+      elem.classList.remove('datetime');
+    }
   });
 };
 
--- a/pkg/web/static/js/main.js
+++ b/pkg/web/static/js/main.js
@@ -66,5 +66,4 @@ const updateDatetimes = (e) => {
   });
 };
 
-document.addEventListener('DOMContentLoaded', updateDatetimes);
 document.body.addEventListener('htmx:afterSettle', updateDatetimes);
--- a/pkg/web/static/js/main.js
+++ b/pkg/web/static/js/main.js
@@ -44,3 +44,25 @@ document.body.addEventListener('htmx:responseError', (e) => {
     }, 5000)
   }
 })
+
+/*
+ * Timestamp handling: localizes and formats datetimes on the page.
+ */
+const updateDatetimes = (e) => {
+  const options = {};
+  const dtfmt = Intl.DateTimeFormat(navigator.languages, {
+    year: 'numeric', month: 'long', day: 'numeric',
+    hour: 'numeric', minute:'numeric', second: 'numeric',
+    hour12: false, timeZoneName: "short",
+  });
+
+  const datetimes = document.querySelectorAll('.datetime');
+  datetimes?.forEach(elem => {
+    const dt = new Date(elem.textContent);
+    elem.textContent = dtfmt.format(dt);
+    elem.classList.remove('datetime');
+  });
+};
+
+document.addEventListener('DOMContentLoaded', updateDatetimes);
+document.body.addEventListener('htmx:afterSettle', updateDatetimes);
--- a/pkg/web/static/js/transactionInfo.js
+++ b/pkg/web/static/js/transactionInfo.js
@@ -30,15 +30,6 @@ document.body.addEventListener('htmx:afterSettle', (e) => {
     const humanizeLastUpdate = dayjs(lastUpdate.textContent).fromNow();
     lastUpdate.textContent = humanizeLastUpdate;
   }
-
-  // Display envelope timestamp in local time.
-  if (e.detail.requestConfig.path === `/v1/transactions/${transactionID}/secure-envelopes` && e.detail.requestConfig.verb === 'get') {
-    const envelopeTimestamp = document.querySelectorAll('.envelope-timestamp');
-    envelopeTimestamp.forEach((timestamp) => {
-      const localTime = dayjs(timestamp.textContent).format('MMM DD, YYYY hh:mm:ss A')
-      timestamp.textContent = localTime
-    })
-  }
 });
 
 // Add code to amend the request parameters before the request is sent.
--- a/pkg/web/templates/partials/account/account_list.html
+++ b/pkg/web/templates/partials/account/account_list.html
@@ -22,7 +22,7 @@
           {{ else }}
           <td class="px-16">-</td>
           {{ end }}
-          {{ $created := .Created.Format "January 2, 2006 15:04:05 -0700" }}
+          {{ $created := .Created.Format "January 2, 2006 15:04:05" }}
           <td>{{ $created }}</td>
           <td>
             <div class="dropdown">
--- a/pkg/web/templates/partials/account/account_list.html
+++ b/pkg/web/templates/partials/account/account_list.html
@@ -22,8 +22,8 @@
           {{ else }}
           <td class="px-16">-</td>
           {{ end }}
-          {{ $created := .Created.Format "January 2, 2006 15:04:05" }}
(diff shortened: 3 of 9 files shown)


pkg/web/static/js/main.js
```diff
@@ -43,4 +43,26 @@ document.body.addEventListener('htmx:responseError', (e) => {
...
+  const datetimes = document.querySelectorAll('.datetime');
+  datetimes?.forEach(elem => {
+    const dt = new Date(elem.textContent);
+    elem.textContent = dtfmt.format(dt);
+    elem.classList.remove('datetime');
+  });
+};
+
+document.addEventListener('DOMContentLoaded', updateDatetimes);
+document.body.addEventListener('htmx:afterSettle', updateDatetimes);
```
bbengfort, line 68: Not sure if this is the right way to perform the conversions.

pkg/web/templates/partials/account/account_list.html
```diff
@@ -22,8 +22,8 @@
           {{ else }}
           <td class="px-16">-</td>
           {{ end }}
-          {{ $created := .Created.Format "January 2, 2006 15:04:05 -0700" }}
-          <td>{{ $created }}</td>
+          {{ $created := .Created.Format "2006-01-02T15:04:05-0700" }}
+          <td class="datetime">{{ $created }}</td>
```
bbengfort, line 26: Basically `updateDatetimes` looks for elements with this class, then converts it to a localized timestamp, removing the class when it's done so that it is not executed multiple times on accident.

pkg/web/static/js/main.js
```diff
@@ -43,4 +43,26 @@ document.body.addEventListener('htmx:responseError', (e) => {
...
   }
-})
\ No newline at end of file
+})
+
+/*
+ * Timestamp handling: localizes and formats datetimes on the page.
+ */
+const updateDatetimes = (e) => {
+  const options = {};
```
bbengfort, line 52: oops, this line needs to be removed … 

```suggestion
```

pkg/web/static/js/main.js
```diff
@@ -43,4 +43,26 @@ document.body.addEventListener('htmx:responseError', (e) => {
...
+    hour12: false, timeZoneName: "short",
+  });
+
+  const datetimes = document.querySelectorAll('.datetime');
+  datetimes?.forEach(elem => {
+    const dt = new Date(elem.textContent);
+    elem.textContent = dtfmt.format(dt);
+    elem.classList.remove('datetime');
+  });
+};
```
bbengfort, line 65: I'm guessing this probably needs some error handling or fallback.

pkg/web/static/js/main.js
```diff
@@ -43,4 +43,26 @@ document.body.addEventListener('htmx:responseError', (e) => {
...
+
+document.addEventListener('DOMContentLoaded', updateDatetimes);
```
daniellemaxwell, line 67: Was `DOMContentLoaded` added here for when content isn't swapped? I believe an `afterSettle` event occurs whenever there's a `GET` request which should be triggered for all of the tables. We can double check though.

pkg/web/static/js/main.js
```diff
@@ -43,4 +43,26 @@ document.body.addEventListener('htmx:responseError', (e) => {
...
-})
\ No newline at end of file
+})
+
+/*
+ * Timestamp handling: localizes and formats datetimes on the page.
+ */
+const updateDatetimes = (e) => {
+  const options = {};
+  const dtfmt = Intl.DateTimeFormat(navigator.languages, {
```
daniellemaxwell, line 53: Would we eventually want use a library like `dayjs` to handle this? We're currently using the CDN version to humanize `Last Update`. It has required installing additional plugins, but it could be easier to deal with once we switch to using the package.

pkg/web/static/js/main.js
bbengfort: Yeh, I didn't know if there were any timestamps in not HTMX areas; I didn't even know if an event handler was the right thing. Happy for your advise on how to handle this.

pkg/web/static/js/main.js
bbengfort: I have no problems switching to dayjs when we can start packing the full javascript -- heck maybe we should also use jQuery and underscore to make coding easier as well! If you think we should use dayjs right now, that's fine; we can close this PR and do that in a follow on.

pkg/web/static/js/main.js
```diff
@@ -43,4 +43,26 @@ document.body.addEventListener('htmx:responseError', (e) => {
...
+    year: 'numeric', month: 'long', day: 'numeric',
+    hour: 'numeric', minute:'numeric', second: 'numeric',
+    hour12: false, timeZoneName: "short",
+  });
+
+  const datetimes = document.querySelectorAll('.datetime');
+  datetimes?.forEach(elem => {
+    const dt = new Date(elem.textContent);
+    elem.textContent = dtfmt.format(dt);
+    elem.classList.remove('datetime');
```
daniellemaxwell, line 63: When a new resource is added (transaction, user, etc.), will the class name appear on the newly created element after the swap event?

pkg/web/static/js/main.js
daniellemaxwell: Yes, I think a fallback would be helpful just in case.

pkg/web/static/js/main.js
daniellemaxwell: I should've noted that question was more for myself. I created an account and user and confirmed that the timestamp updated as expected.

pkg/web/static/js/main.js
daniellemaxwell: We only have timestamps in the tables so all of those are swapped in with htmx. I escaped this line and the timestamps still appeared in the correct time zone. This line can be removed, but feel free to double check on your end.

pkg/web/static/js/main.js
daniellemaxwell: We can go ahead and use dayjs. I was thinking of waiting until we start using a bundler, but I can create a story to use dayjs now.

bbengfort: Ok, if you're going to use dayjs - let's just go ahead and close this PR.

pkg/web/static/js/main.js
```diff
@@ -43,4 +43,27 @@ document.body.addEventListener('htmx:responseError', (e) => {
...
+  const options = {};
+  const dtfmt = Intl.DateTimeFormat(navigator.languages, {
+    year: 'numeric', month: 'long', day: 'numeric',
+    hour: 'numeric', minute:'numeric', second: 'numeric',
+    hour12: false, timeZoneName: "short",
+  });
+
+  const datetimes = document.querySelectorAll('.datetime');
+  datetimes?.forEach(elem => {
+    if (elem.textContent !== '' || elem.textContent !== null) {
```
daniellemaxwell, line 61: This check was added to ensure the `textContent` has a timestamp.

pkg/web/static/js/transactionInfo.js
```diff
@@ -30,15 +30,6 @@ document.body.addEventListener('htmx:afterSettle', (e) => {
     const humanizeLastUpdate = dayjs(lastUpdate.textContent).fromNow();
     lastUpdate.textContent = humanizeLastUpdate;
   }
-
-  // Display envelope timestamp in local time.
-  if (e.detail.requestConfig.path === `/v1/transactions/${transactionID}/secure-envelopes` && e.detail.requestConfig.verb === 'get') {
```
daniellemaxwell, line 35: This was removed since the time zone wasn't displayed and we would've encountered the bug I discovered about the offset appearing instead of the time zone. Also, the timestamp should be consistent with others in the UI.

pkg/web/templates/partials/secure_envelope/secure_envelope_list.html
```diff
@@ -5,11 +5,11 @@
   {{ $sentAt := .SentAt.Format "January 2, 2006 3:04:05 PM" }}
   {{ $receivedAt := .ReceivedAt.Format "January 2, 2006 3:04:05 PM" }}
   <p>
-    <span class="font-semibold">Sent at:</span>
+    <span class="datetime font-semibold">Sent at:</span>
```
daniellemaxwell, line 8: Updating the format for timestamps in the secure envelope list.

pkg/web/static/js/main.js
```diff
@@ -43,4 +43,26 @@ document.body.addEventListener('htmx:responseError', (e) => {
...
+      const dt = new Date(elem.textContent);
+      elem.textContent = dtfmt.format(dt);
+      elem.classList.remove('datetime');
+    }
+  });
+};
+
+document.body.addEventListener('htmx:afterSettle', updateDatetimes);
```
daniellemaxwell, line 68: The `DOMContentLoaded` event listener has been removed since all of the timestamps are populated after a swap event.

daniellemaxwell (approved):
Using the Intl.DateTimeFormat is a great way to handle the timestamp formatting. A similar approach was used in Ensign.

I updated the timestamp in the secure envelope list and made a few minor changes to the main.js file. If you have any questions, let me know.

PR #170, opened 2024-07-21
Add Envelope Decryption to API

bbengfort:
### Scope of changes

Decrypts envelopes on request in SecureEnvelope List and Detail API handlers.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

There are a few things that still need to be taken care of, including adding the remote peer to the list return and using that common name in the lookup of the unsealing keys.

### Author checklist

### Reviewer(s) checklist

- [ ] The data returned in the API contains all required fields for the user interface

### DIFF
--- a/pkg/web/transactions.go
+++ b/pkg/web/transactions.go
@@ -325,7 +325,7 @@ func (s *Server) AcceptTransactionPreview(c *gin.Context) {
 		return
 	}
 
-	if out, err = api.NewEnvelope(decrypted); err != nil {
+	if out, err = api.NewEnvelope(env, decrypted); err != nil {
 		c.Error(err)
 		c.JSON(http.StatusInternalServerError, api.Error(err))
 		return
@@ -349,7 +349,8 @@ func (s *Server) SendEnvelopeForTransaction(c *gin.Context) {
 		db           models.PreparedTransaction
 		payload      *trisa.Payload
 		outgoing     *envelope.Envelope
-		incoming     *envelope.Envelope
+		incoming     *models.SecureEnvelope
+		decrypted    *envelope.Envelope
 	)
 
 	ctx := c.Request.Context()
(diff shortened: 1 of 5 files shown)


pkg/web/transactions.go
```diff
@@ -455,10 +457,24 @@ func (s *Server) SendEnvelopeForTransaction(c *gin.Context) {
...
-	// TODO: we need to get the incoming envelope in order for this to work!
+	// Retrieve the secure envelope model for the incoming envelope
+	if incoming, err = s.store.LatestSecureEnvelope(ctx, transaction.ID, models.DirectionIncoming); err != nil {
+		c.Error(fmt.Errorf("could not retrieve incoming secure envelope: %w", err))
+		c.JSON(http.StatusInternalServerError, api.Error("could not return incoming response from counterparty"))
+		return
+	}
+
+	// Decrypt the incoming secure envelope
+	// TODO: why are we decrypting the incoming secure envelope again?
```
daniellemaxwell, line 468: Are you asking if the secure envelope should be decrypted?

pkg/web/transactions.go
```diff
@@ -560,16 +577,35 @@ func (s *Server) RejectTransaction(c *gin.Context) {
...
-		Data:     out,
-		HTMLName: "transaction_reject.html",
-	})
+	detailURL, _ := url.JoinPath("/transactions", transaction.ID.String(), "info")
+	// Set a cookie to show a toast message on the page redirect.
+	setToastCookie(c, "transaction_reject_success", "true", detailURL, s.conf.Auth.CookieDomain)
+
+	// If the content requested is HTML (e.g. the web-front end), then redirect the user
+	// to the transaction detail page.
+	if c.NegotiateFormat(binding.MIMEJSON, binding.MIMEHTML) == binding.MIMEHTML {
```
daniellemaxwell, line 586: Do we need to redirect users if a rejection request is successful? The reject modal appears when users click the `Reject` button the detail page, so they're all ready at the detail page. Once the modal is closed and if the request is successful, we should be able to display a toast message without needing to redirect them.

Users may also access the modal on the transaction list page. If we only want them to view the modal on the detail page, the UI may be updated to not display the reject modal on the transaction list page.

pkg/web/transactions.go
```diff
@@ -648,7 +697,7 @@ func (s *Server) SecureEnvelopeDetail(c *gin.Context) {
 		transactionID uuid.UUID
 		envelopeID    ulid.ULID
 		model         *models.SecureEnvelope
-		out           *api.SecureEnvelope
+		out           any
```
daniellemaxwell, line 700: Does the change to `any` here mean that the FE will have to check if the return is an `Envelope` or a `SecureEnvelope` to display the envelope detail data?

pkg/web/transactions.go
bbengfort: No, it's just that it's the second time the decryption is happening; the first time when it came into the node, then was saved to the database (encrypted) then we decrypt it again after loading it from the database; it's wasted extra work, but I'd have to be able to access the incoming message in memory somehow to not do the second decryption.

I think this means creating a better sending mechanism that wraps the various functionality required.

pkg/web/transactions.go
bbengfort: I was following the same pattern as Accept, but certainly, we can do things differently for reject. The reason for the redirect is because we can't simply re-render the transaction detail page; but if you just need an AJAX 204 response back we can make that happen (or some other partial, is there a swap happening?)

pkg/web/transactions.go
bbengfort: Yes it will - though the front-end does have control of whether or not it is asking for decrypted envelopes or encrypted secure envelopes with the query string. I tried to make the basic details of both the Envelope and the SecureEnvelope (such as ID, EnvelopeID, direction, etc) the same though to reduce friction.

pkg/web/transactions.go
daniellemaxwell: That makes sense. I saw this first and then saw this comment elsewhere and wanted to make sure I understood the thought process.

daniellemaxwell (approved):
This looks good! Thank you for making these changes and for explaining follow on work that will be required.

pkg/web/transactions.go
bbengfort: Ok, I'll work on this in SC-25929

PR #172, opened 2024-07-22
Use input event on password and confirm password

daniellemaxwell:
### Scope of changes

This PR amends the event listeners added to the `password` and `confirmation password` input elements.

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [x] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/29779677?key=421f6a4f14b89374d0db8acfd8a2c076

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/web/static/js/usersManagement.js
+++ b/pkg/web/static/js/usersManagement.js
@@ -53,7 +53,6 @@ document.addEventListener('htmx:afterSettle', (e) => {
       };
     };
 
-    // TODO: Should input also be included in the events lists?
     addEventListeners(passwordInput, 'input', checkPasswords);
     addEventListeners(confirmPassword, 'input', checkPasswords);
   };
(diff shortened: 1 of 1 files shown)


pkg/web/static/js/usersManagement.js
```diff
@@ -54,8 +54,8 @@ document.addEventListener('htmx:afterSettle', (e) => {
     };
 
     // TODO: Should input also be included in the events lists?
-    addEventListeners(passwordInput, 'change keyup paste cut', checkPasswords);
-    addEventListeners(confirmPassword, 'change keyup paste cut', checkPasswords);
+    addEventListeners(passwordInput, 'input', checkPasswords);
```
daniellemaxwell, line 57: I double checked whether `input` would also be needed here. I determined that adding `input` means that the others aren't required since the input is fired every time a change is made to the input, so this would cover 'keyup' and 'paste'. Since we're using the `password` type for the element, users aren't currently able to `cut`. 

The `change` event on the other hand only fires once a value has been committed.

bbengfort: glad you found a better event!

bbengfort (approved):
TIL!

PR #181, opened 2024-07-30
Dynamic OpenAPI

bbengfort:
### Scope of changes

This PR makes the `openapi.json` and `openapi.yaml` specifications a template to dynamically render per-node data; e.g. so that users can use their API documentation to make requests to their own node.

https://www.awesomescreenshot.com/video/30038462?key=868dc30d097787fd417056b24b35e465

### Type of change

- [ ] bug fix
- [x] new feature
- [x] documentation
- [ ] other (describe)

### Acceptance criteria

This was a request from a user so they didn't have to use postman or another tool in order to test against the API.

Questions:

1. Should we add a better server description? E.g. Rotational Envoy or Rotational TestNet Envoy? 
2. Are there any variables we should add to the server definition?
3. Is this going to make the API documentation impossible to maintain?
4. Are there any other fields that need to be dynamic?

### Author checklist

### Reviewer(s) checklist

- [ ] Maintaining this documentation will not be too much of a pain in the butt even though we'll have to make sure the Go template variables are correctly added.


pkg/web/openapi.go
```diff
@@ -0,0 +1,56 @@
...
+		data       gin.H
+		templates  *template.Template
+		initialize sync.Once
+	)
+
+	initialize.Do(func() {
+		data = gin.H{
+			"Version":     pkg.Version(),
+			"Origin":      s.conf.Web.Origin,
+			"Description": "Envoy",
```
daniellemaxwell, line 28: Using `Rotational Envoy` here sounds fine to me. It would be a bit more specific than just `Envoy`.

pkg/web/templates/openapi/openapi.json
```diff
@@ -11,12 +11,19 @@
...
+    "version": "v{{ .Version }}"
   },
   "externalDocs": {
     "description": "TRISA Developer Documentation",
     "url": "https://trisa.dev"
   },
+  "servers": [
+    {
+      "url": "{{ .Origin }}",
+      "variables": {},
```
daniellemaxwell, line 23: I can't think of any `variables` that should be added here. Then again, I don't use the API docs a whole lot so there could be something others would find more useful that I'm not thinking of right now.

daniellemaxwell (approved):
This is will be another way to reduce friction for users, so this looks should be very helpful. I'm not sure if this will cause issues in terms of maintaining the docs. Though I would imagine we'd want to test that changes made to the docs work on the documentation site as expected. If you'd like to discuss the questions you have about this one further, let me know.

PR #183, opened 2024-07-31
Update statuses and display tooltip

daniellemaxwell:
### Scope of changes

This PR refactors how the transaction status color is applied. It adds a tooltip to explain each status to users on hover. Also, svg icons have been replaced with a font awesome icon.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/30073636?key=fe040db0933814a963d1d507489af1b1

### Author checklist

### Reviewer(s) checklist

- [ ] Is each transaction status included in `getStatus`?


pkg/web/static/styles/style.css
```diff
@@ -75,4 +75,9 @@ body main {
 
 .htmx-request.htmx-indicator{
   display: inline;
+}
+
+/* Tooltip styles */
+#status-tooltip:after {
```
daniellemaxwell, line 81: The tooltip wasn't centered above the text content. This change resolved that.

tailwind.config.js
```diff
@@ -8,6 +8,13 @@ module.exports = {
     extend: {},
   },
   plugins: [require("daisyui")],
+  safelist: [
```
daniellemaxwell, line 11: The text color classes are now applied dynamically. To ensure that the colors are applied as expected in the browser, they have been added here to Tailwind's safe list.

pkg/web/api/v1/transaction.go
```diff
@@ -221,10 +221,94 @@ func (c *Transaction) Model() (model *models.Transaction, err error) {
...
 
+func (c *Transaction) ColorStatus() string {
+	return getStatus(c.Status, color)
+}
+
+func (c *Transaction) TooltipStatus() string {
+	return getStatus(c.Status, tooltip)
+}
+
+func getStatus(status, attr string) string {
```
bbengfort, line 245: I like the idea here to unify handling in a single `getStatus` method; having the color and the tool tip in the same function just a few lines apart makes it easier to edit them together. In a compiled language like Go, however, the nested if statements inside of the switch will prevent any compiler optimization from occurring. This is one of the reasons Go is so verbose; because we're trying to make maximal use of the CPU and be as fast as possible, the code allows us to to take advantage of the compiler optimizations now and in the future. 

So a function like `getStatus` is better in Python, Ruby, R, and Javascript than it is in C, Go, Rust, or Java. 

I'll submit a commit that does the refactoring and uses constants to do what you intended; make the editing of the statuses a bit easier; it's just that there will be two switch statements instead of one.

bbengfort (approved):
The tooltips and colors make the UX so much more friendly; thank you for adding this! I hope you don't mind, I did a little bit of refactoring to make the Go code a bit more performant but still keeping the spirt of ease of maintenance that you intended.

pkg/web/api/v1/transaction.go
daniellemaxwell: I initially had two separate switch statements, but then thought it felt redundant; so I really appreciate this explanation.

pkg/web/api/v1/transaction.go
```diff
@@ -226,87 +226,72 @@ func (c *Transaction) Model() (model *models.Transaction, err error) {
...
+	case models.StatusRepair:
+		return colorRepair
+	case models.StatusAccepted:
+		return colorAccepted
+	case models.StatusCompleted:
+		return colorCompleted
+	case models.StatusRejected:
+		return colorRejected
+	default:
+		panic(fmt.Errorf("unhandled color for status %q", c.Status))
```
daniellemaxwell, line 272: I like the inclusion of the panic.

PR #191, opened 2024-08-06
Fix Status Bugs

bbengfort:
### Scope of changes

This PR makes fixes to how the transaction state is modified in order to ensure it matches the expected Envoy workflows.

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

The best way to review this PR is to run locally and go through the workflows to make sure they match. I did do this manually.

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/postman/outgoing.go
+++ b/pkg/postman/outgoing.go
@@ -211,6 +211,51 @@ func (o *Outgoing) reseal(storageKey keys.PublicKey, sec crypto.Crypto) (err err
 // message transfer state. For example, if the outgiong transfer state is pending, then
 // the Transfer should be marked as needing review.
 func (o *Outgoing) StatusFromTransferState() string {
+	if o.packet.Reply == DirectionOutgoing {
+		// If this is a reply to an incoming packet and the response is Pending, then
+		// we need to determine what state the incoming message was in to determine
+		// what pending action needs to be completed.
+		switch ts := o.Envelope.TransferState(); ts {
+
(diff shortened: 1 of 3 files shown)


pkg/postman/outgoing.go
```diff
@@ -211,6 +211,51 @@ func (o *Outgoing) reseal(storageKey keys.PublicKey, sec crypto.Crypto) (err err
...
+			default:
+				panic(fmt.Errorf("unhandled incoming transfer state %q to determine transaction status after pending reply", ints.String()))
+			}
+
+		// If we're sending review or repair, we're waiting for the recipient to take action
+		case api.TransferReview, api.TransferRepair:
+			return models.StatusPending
+
+		// These are the echo back statuses; we assume the incoming message matches.
+		case api.TransferAccepted:
```
daniellemaxwell, line 246: Would there need to be another state for the transaction is `accepted` but the `Transaction ID` needs to be sent in order for the state to become `complete`?

daniellemaxwell (approved):
The transfer states updated as expected when I tested locally. This should make it easier to implement the repair flow. Thanks for making these updates so quickly!

pkg/postman/outgoing.go
bbengfort: My idea was to leave it as accepted -- that would be the indication that the originator needs to complete it. I think accepted does imply the action of "go ahead with the on-chain transfer"

PR #199, opened 2024-08-13
Sc 26971/api page

daniellemaxwell:
### Scope of changes

This PR adds the api key management page and the create api key modal.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/30477717?key=7e98fc5ac88c8e6322a3947ef4006e2f

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/web/routes.go
+++ b/pkg/web/routes.go
@@ -104,8 +104,8 @@ func (s *Server) setupRoutes() (err error) {
 	s.router.GET("/accounts", authenticate, s.Accounts)
 	s.router.GET("/counterparty", authenticate, s.CounterpartyVasps)
 	s.router.GET("/users", authenticate, s.UsersManagement)
-	s.router.GET("/utilities/travel-address", authenticate, s.TravelAddressUtility)
 	s.router.GET("/apikeys", authenticate, s.APIKeys)
+	s.router.GET("/utilities/travel-address", authenticate, s.TravelAddressUtility)
 
 	// Swagger documentation with Swagger UI hosted from a CDN
 	s.router.GET("/v1/docs/openapi.:ext", s.OpenAPI)
--- a/pkg/web/routes.go
+++ b/pkg/web/routes.go
@@ -105,6 +105,7 @@ func (s *Server) setupRoutes() (err error) {
 	s.router.GET("/counterparty", authenticate, s.CounterpartyVasps)
 	s.router.GET("/users", authenticate, s.UsersManagement)
 	s.router.GET("/utilities/travel-address", authenticate, s.TravelAddressUtility)
+	s.router.GET("/apikeys", authenticate, s.APIKeys)
 
 	// Swagger documentation with Swagger UI hosted from a CDN
 	s.router.GET("/v1/docs/openapi.:ext", s.OpenAPI)
--- a/pkg/web/templates/components/add_apikey.html
+++ b/pkg/web/templates/components/add_apikey.html
@@ -26,8 +26,7 @@
         <section class="p-2 my-2 border border-gray-300 rounded">
           <h3 class="font-bold">Full Access</h3>
           <div class="flex justify-between">
-            <label for="full_access" class="max-w-4xl text-balance">(Manage accounts, View accounts, Manage api keys, View api keys, Revoke api keys, Manage config, View config, Manage counterparties, View counterparties, 
-              Manage pki, View pki, Delete pki, Manage transactions, View transactions, Manage travel rule, View travel rule, Delete travel rule, Manage users, View users)
+            <label for="full_access" class="max-w-4xl text-balance">(Manages accounts, api keys, configuration, counterparties, PKI, travel rule interactions, and users.)
             </label>
             <div class="pt-1">
               <input type="checkbox" checked id="full_access" name="permissions" value="full" class="checkbox-style" />
@@ -88,14 +87,6 @@
           <div class="my-2 grid grid-cols-2 items-center">
             <label for="pki_delete">pki:delete</label>
             <input type="checkbox" id="pki_delete" name="permissions[]" value="pki:delete" class="checkbox-style" />
-          </div>
-          <div class="my-2 grid grid-cols-2 items-center">
-            <label for="transactions_manage">transactions:manage</label>
-            <input type="checkbox" id="transactions_manage" name="permissions[]" value="transactions:manage" class="checkbox-style" />
-          </div>
-          <div class="my-2 grid grid-cols-2 items-center">
-            <label for="transactions_view">transactions:view</label>
-            <input type="checkbox" id="transactions_view" name="permissions[]" value="transactions:view" class="checkbox-style" />
           </div>
           <div class="my-2 grid grid-cols-2 items-center">
             <label for="travelrule:manage">travelrule:manage</label>
--- a/pkg/web/templates/components/add_apikey.html
+++ b/pkg/web/templates/components/add_apikey.html
@@ -1,0 +1,130 @@
+<dialog id="add_apikey_modal" class="modal">
+  <div class="modal-box max-w-2xl">
+    <div class="flex justify-between items-center">
+      <h1 class="font-bold text-xl">Create API Key</h1>
+      <button onclick="add_apikey_modal.close()" class="btn btn-sm btn-circle btn-ghost">
+        <i class="fa-solid fa-x"></i>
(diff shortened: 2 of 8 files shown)


pkg/web/templates/components/add_apikey.html
```diff
@@ -0,0 +1,130 @@
...
+        >
+        <div class="mb-4">
+          <label for="description" class="label-style">API Key Description</label>
+          <input type="text" id="description" name="description" class="input-style" />
+        </div>
+        <h2 class="font-bold">Permissions:</h2>
+        <section class="p-2 my-2 border border-gray-300 rounded">
+          <h3 class="font-bold">Full Access</h3>
+          <div class="flex justify-between">
+            <label for="full_access" class="max-w-4xl text-balance">(Manage accounts, View accounts, Manage api keys, View api keys, Revoke api keys, Manage config, View config, Manage counterparties, View counterparties,
```
daniellemaxwell, line 29: Are the permissions correct?

pkg/web/templates/components/add_apikey.html
```diff
@@ -0,0 +1,130 @@
...
+            <div class="pt-1">
+              <input type="checkbox" checked id="full_access" name="permissions" value="full" class="checkbox-style" />
+            </div>
+          </div>
+        </section>
+        <section class="p-2 my-4 border border-gray-300 rounded">
+          <div class="flex justify-between">
+            <label for="custom_access" class="font-bold">Custom Access</label>
+            <div class="pt-1">
+              <input type="checkbox" id="custom_access" class="checkbox-style" />
```
daniellemaxwell, line 41: Do we want to hide the permissions and then display them once a users clicks custom access? If yes, the custom permissions will be hidden and once the user clicks this checkbox they will appear.

pkg/web/templates/components/add_apikey.html
```diff
@@ -0,0 +1,130 @@
...
+        </div>
+        <h2 class="font-bold">Permissions:</h2>
+        <section class="p-2 my-2 border border-gray-300 rounded">
+          <h3 class="font-bold">Full Access</h3>
+          <div class="flex justify-between">
+            <label for="full_access" class="max-w-4xl text-balance">(Manage accounts, View accounts, Manage api keys, View api keys, Revoke api keys, Manage config, View config, Manage counterparties, View counterparties, 
+              Manage pki, View pki, Delete pki, Manage transactions, View transactions, Manage travel rule, View travel rule, Delete travel rule, Manage users, View users)
+            </label>
+            <div class="pt-1">
+              <input type="checkbox" checked id="full_access" name="permissions" value="full" class="checkbox-style" />
```
daniellemaxwell, line 33: If the user selects full then I'll use the `htmx:configRequest` event to update the params to send the full list of permissions.

pkg/web/routes.go
```diff
@@ -105,6 +105,7 @@ func (s *Server) setupRoutes() (err error) {
 	s.router.GET("/counterparty", authenticate, s.CounterpartyVasps)
 	s.router.GET("/users", authenticate, s.UsersManagement)
 	s.router.GET("/utilities/travel-address", authenticate, s.TravelAddressUtility)
+	s.router.GET("/apikeys", authenticate, s.APIKeys)
```
bbengfort, line 108:
```suggestion
	s.router.GET("/users", authenticate, s.UsersManagement)
	s.router.GET("/apikeys", authenticate, s.APIKeys)
	s.router.GET("/utilities/travel-address", authenticate, s.TravelAddressUtility)
```

Just for organization purposes, could we keep things in the order that the routes are defined; in this case users, apikeys, utilities?

pkg/web/templates/components/add_apikey.html
bbengfort: I wouldn't worry about having all of the permissions in the label; especially since we might be adding permissions in the future. Perhaps just change the label to something like the suggestion?

pkg/web/templates/components/add_apikey.html
```diff
@@ -0,0 +1,130 @@
...
+        <div class="mb-4">
+          <label for="description" class="label-style">API Key Description</label>
+          <input type="text" id="description" name="description" class="input-style" />
+        </div>
+        <h2 class="font-bold">Permissions:</h2>
+        <section class="p-2 my-2 border border-gray-300 rounded">
+          <h3 class="font-bold">Full Access</h3>
+          <div class="flex justify-between">
+            <label for="full_access" class="max-w-4xl text-balance">(Manage accounts, View accounts, Manage api keys, View api keys, Revoke api keys, Manage config, View config, Manage counterparties, View counterparties, 
+              Manage pki, View pki, Delete pki, Manage transactions, View transactions, Manage travel rule, View travel rule, Delete travel rule, Manage users, View users)
```
bbengfort, line 30:
```suggestion
            <label for="full_access" class="max-w-4xl text-balance">(Manages accounts, api keys, configuration, counterparties, PKI, travel rule interactions, and users.)
```

pkg/web/templates/components/add_apikey.html
bbengfort: Using Javascript to get the full list from the permissions input? E.g. is all we have to manage the checkbox list to add/remove permissions?

pkg/web/templates/components/add_apikey.html
bbengfort: No, I like having the permissions listed by default, so long as the vertical scroll doesn't hide the submit button (e.g. the permissions list should scroll independently of the rest of the modal).

pkg/web/templates/components/add_apikey.html
```diff
@@ -0,0 +1,130 @@
...
+            <input type="checkbox" id="pki_delete" name="permissions[]" value="pki:delete" class="checkbox-style" />
+          </div>
+          <div class="my-2 grid grid-cols-2 items-center">
+            <label for="transactions_manage">transactions:manage</label>
+            <input type="checkbox" id="transactions_manage" name="permissions[]" value="transactions:manage" class="checkbox-style" />
+          </div>
+          <div class="my-2 grid grid-cols-2 items-center">
+            <label for="transactions_view">transactions:view</label>
+            <input type="checkbox" id="transactions_view" name="permissions[]" value="transactions:view" class="checkbox-style" />
+          </div>
```
bbengfort, line 99: There are no transactions permissions, the travelrule permission manages transactions and secure envelopes. 

```suggestion
```

pkg/web/templates/components/add_apikey.html
```diff
@@ -0,0 +1,130 @@
...
+          <div class="my-2 grid grid-cols-2 items-center">
+            <label for="users:view">users:view</label>
+            <input type="checkbox" id="users:view" name="permissions[]" value="users:view" class="checkbox-style" />
+          </div>
+        </section>
+        <div class="flex justify-center">
+          <button id="acct-submit-bttn" class="submit-btn w-44 btn bg-primary font-semibold text-lg text-white hover:bg-primary/90">
+            <span class="submit-btn-text">Create</span>
+            <span id="loader" class="htmx-indicator loading loading-spinner loading-md"></span>
+          </button>
```
bbengfort, line 125: I think that's all the permissions! The list is here:

https://github.com/trisacrypto/envoy/blob/main/pkg/web/auth/permissions/permissions.go#L45-L54

There should be 18 including "unknown" which can be omitted from this list, so 17.

bbengfort (approved):
This looks really great! I'll be very happy not to have to manage API keys manually!

pkg/web/templates/components/add_apikey.html
daniellemaxwell: I'll make the update while working on the delete modal view.

pkg/web/templates/components/add_apikey.html
daniellemaxwell: Thanks! I think I originally typed `transactions` and instead of changing it to travel rule I ended up keeping both.

pkg/web/templates/components/add_apikey.html
daniellemaxwell: If the user selects full, then would we need to send an array with all of the permissions? I'm under the impression that `"permissions": "full"` would not be accepted, so the param would need to be updated w/ JS to send all of the permissions inside an array.

PR #203, opened 2024-08-16
Replace actions menu with button(s)

daniellemaxwell:
### Scope of changes

This PR replaces the actions menu with buttons to allow users to trigger specific actions depending on the resource.

Making the table rows clickable was previously discussed, but that change would make it so that users aren't able to copy data displayed in a table. Also, the change would've either had to be made with JS or by adding a link to each `<td> `. Other options would've required significant CSS changes or not using `<table>` which then requires having to make the element replacing the `table` accessible.

This change would be consistent with the `Delete` button currently displayed in the `api keys` table.

Questions:
- Do we want to use Action buttons instead of making the table rows clickable?
- If we move forward with the button, do we still want to include a preview button in the transactions list table?

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [x] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/30573463?key=8a81008160c7fd306955d4791397fa90

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/web/templates/partials/counterparty/counterparty_list.html
+++ b/pkg/web/templates/partials/counterparty/counterparty_list.html
@@ -15,7 +15,7 @@
     <tbody>
       {{ if .Counterparties }}
       {{ range .Counterparties }}
-      <tr class="hover">
+      <tr>
         <td>{{ .Name }}</td>
         <td>{{ .Website }}</td>
         <td>{{ .Protocol }}</td>
--- a/pkg/web/templates/partials/counterparty/counterparty_list.html
+++ b/pkg/web/templates/partials/counterparty/counterparty_list.html
@@ -1,7 +1,7 @@
 {{- $canEditAccounts := not .IsViewOnly -}}
 {{- with .CounterpartyList -}}
 <div class="table-container pb-36 overflow-x-auto overflow-y-hidden">
-  <table class="table table-sm">
+  <table class="table">
     <thead class="font-bold text-base text-black">
       <tr>
         <th>Name</th>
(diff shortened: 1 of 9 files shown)


pkg/web/templates/partials/counterparty/counterparty_list.html
```diff
@@ -1,36 +1,36 @@
...
-            <div tabindex="0" role="button" class="btn btn-ghost font-bold text-xl">&hellip;</div>
-            <ul tabindex="0" class="dropdown-content z-[1] menu p-2 shadow bg-black font-semibold text-white rounded-box">
-              {{ if and (eq .Source "user") ($canEditAccounts) }}
-              <li><a onclick="cparty_modal.showModal()" hx-get="/v1/counterparties/{{ .ID }}/edit" hx-target="#cparty_modal" hx-swap="innerHTML">Edit</a></li>
-              {{ end }}
-              <li><a>View</a></li>
-            </ul>
+          <div class="flex gap-x-2 items-center">
+            {{ if and (eq .Source "user") ($canEditAccounts) }}
+            <button type="button" class="btn btn-sm bg-primary text-white hover:bg-primary/90" onclick="cparty_modal.showModal()"
```
daniellemaxwell, line 28: I'd like to note that I added the `hx-indicator` attribute here to display a loader component, but it didn't work. I'll create a follow on story to figure that out.

pdeziel (approved):
Ok so the question is whether to make the rows clickable or have an actions button? I'm leaning on the side of the button. We could also use the ellipses thing (...) if the problem is the button is too clunky.

daniellemaxwell:
@pdeziel Yes, the main question was about whether we should use the button instead. The other one I had can be discussed during a future sync.

PR #215, opened 2024-08-27
Add update api key modal and implement functionality

daniellemaxwell:
### Scope of changes

This PR adds an update API key modal that will allow users to change a key's description. A minor change was also made to display the key's description in the revoke API key modal.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/30908152?key=b7038e74ea08e270033369ce34fcca28

### Author checklist

### Reviewer(s) checklist


pkg/web/api/v1/apikey.go
```diff
@@ -27,6 +32,10 @@ type APIKeyList struct {
 	APIKeys []*APIKey  `json:"api_keys"`
 }
 
+type APIKeyQuery struct {
```
daniellemaxwell, line 35: A query has been added so that we'll know which modal should be displayed to the user.

bbengfort (commented):
Have you heard of the term "[code smell](https://martinfowler.com/bliki/CodeSmell.html)"? I'm sensing that something is a little off with this implementation but I can't quite point to the problem that I think it might lead to. 

First, I just want to check my understanding of the GET query: when the user clicks the "Edit" or "Revoke" button an htmx query goes out in order to fetch the HTML data to populate the modal, which before the htmx query is just an empty div. The query flag is used to tell the detail endpoint what kind of modal to render. Is that correct?

I guess something seems off about this since the detail workflow is tied to the revoke workflow and they should be independent. It would also imply that if this was required for other objects like deleting counterparties there would need to be another query flag for preview (or that's just the default rendering). The endpoint is doing a little too much work. 

What the revoke modal needs is the endpoint to send the DELETE request too. I propose that we put that endpoint in a `data-endpoint` attribute of the button itself. When the user clicks the revoke button, we can open the revoke modal entirely in Javascript, and because the event has the parent element that was clicked, we can get the endpoint off of the button in order to direct the revoke request. That would prevent the need for a query string and HTMX. 

Also, this is a relatively minor thing, but I'm not sure I like the two text input modal for editing the description of the API key. I think one text box with the original description is simpler and more maintainable.

daniellemaxwell:
@bbengfort Yes, we were retrieving the API key to get the ID for the endpoint and the description to display in the revoke modal. I've made an update so that the endpoint and description are added via JavaScript. We want to keep the `hx-delete` attribute so that it will then trigger the update to the table once a key is revoked. 

Here's a link to a video testing the changes: https://www.awesomescreenshot.com/video/30958427?key=a9814c317bc8331f9df4156453ac173b

bbengfort (approved):

PR #217, opened 2024-08-28
Update counterparty details

daniellemaxwell:
### Scope of changes

This PR gives users the ability to update a counterparty.

Note: I encountered a bug where the country didn't always appear in the modal, after I'd previously opened one, but never saw any error messages. I attempted to debug, but still couldn't sort out the issue. This didn't always happen, so that made it challenging to sort out. I'm hoping it's due to the servers being slow and causing a loading issue here. However, I can't confirm that without testing in different environment.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/30950506?key=9c01493c0ae6c31e8df940a4b7746345

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/web/static/js/counterparties.js
+++ b/pkg/web/static/js/counterparties.js
@@ -16,13 +16,10 @@ document.body.addEventListener('htmx:afterRequest', (e) => {
 });
 
 document.body.addEventListener('htmx:afterRequest', (e) => {
-  const editCpartyForm = 'edit-cparty-form';
-  // Check if the request to update a user added counterparty was successful.
-  if (e.detail.elt.id === editCpartyForm && e.detail.requestConfig.verb === 'put' && e.detail.successful) {
-    // Close the edit counterparty modal and reset the form.
+  // Close the edit counterparty modal after a successful request.
+  if ( e.detail.requestConfig.verb === 'put' && e.detail.successful) {
     document.getElementById('cparty_modal').close();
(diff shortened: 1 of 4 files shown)


bbengfort (approved):
Do you know if counterparties that were created by the directory sync can be modified? I don't remember if I put any safeguards into the API or not. 

The problem is that if they can, then they will revert back to the state from the GDS the next time a sync runs (every 6 hours by default). 

We may have to do something where you check the `source` of the counterparty. If the source is "user" they can edit the counterparty, however if the sources is "gds" then the edit button should not be available.

daniellemaxwell:
@bbengfort In the [counterparty list partial](https://github.com/trisacrypto/envoy/blob/17341cdda8ea616e25e17cf3d82b29512e33e464/pkg/web/templates/partials/counterparty/counterparty_list.html#L27), we display the edit button if the source is `user` and the user has edit permission.

bbengfort:
> @bbengfort In the [counterparty list partial](https://github.com/trisacrypto/envoy/blob/17341cdda8ea616e25e17cf3d82b29512e33e464/pkg/web/templates/partials/counterparty/counterparty_list.html#L27), we display the edit button if the source is `user` and the user has edit permission.

Awesome! Glad we already did that!

PR #232, opened 2024-10-31
Add resend confirmation link view

daniellemaxwell:
### Scope of changes

This PR adds the resend link view for when users click on a a sunrise message link that has expired. If the message hasn't been accepted or completed, but the link has expired users will have the ability to request a new confirmation link to initiate the workflow to view the data.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/image/51353636?key=34bc161f457c32db0185ae1e990831e6

### Author checklist

### Reviewer(s) checklist


pkg/web/templates/partials/sunrise/verification.html
```diff
@@ -1,17 +1,34 @@
-<div id="verify-code" class="mx-auto max-w-4xl">
+<!-- If the confirmation link is valid and hasn't expired, display the email verification -->
```
daniellemaxwell, line 1: This is how I believe the workflow will work. A user will click a link and htmx will trigger this partial to load. If the confirmation link is valid and hasn't expired users will see the verify email form. If the link has expired, the user see the resend confirmation link view.

pkg/web/templates/partials/sunrise/verify_code.html
```diff
@@ -0,0 +1,17 @@
+<div id="verify-code" class="mx-auto max-w-4xl">
```
daniellemaxwell, line 1: Once a user clicks the submit button in the verify email form, this partial would be loaded into the empty div in verify.html.

pkg/web/static/brokenlink.svg
bbengfort, line 1: :-) nice icon!

pkg/web/templates/partials/sunrise/verification.html
bbengfort: That makes sense to me! We can either render a different partial in the back-end or use HX-Redirect, but either way, the front-end stays the same.

bbengfort (approved):
Looks great! I get where you're going with the workflow and it makes sense to me!

PR #235, opened 2024-11-01
Create sunrise not found page

daniellemaxwell:
### Scope of changes

This PR adds a not found page that encourages sunrise users to join TRISA.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Note: The image below display the error page with the error layout and not the layout that has been created for sunrise pages in a separate PR that's currently pending review.

https://tinyurl.com/25mxqyjb

### Author checklist

### Reviewer(s) checklist


pkg/web/templates/sunrise_404.html
```diff
@@ -0,0 +1,15 @@
+{{ template "sunrise" . }}
+{{ define "content" }}
+
+<section class="text-center py-12">
+  <img src="/static/404error.svg" alt="404 Error" class="w-[350px] mx-auto" />
+  <p class="py-4 text-xl">Sorry, we couldn't find the page you're looking for. Try going back to a previous page or contact <a href="mailto:support@rotational.io" class="text-blue-700">support@rotational.io </a>for assistance.</p>
+  <p class="py-4 text-xl">If you're not yet part of the TRISA network, consider joining to streamline future Travel Rule compliance and secure data exchanges.</p>
+
+  <div class="my-10 flex items-center gap-x-4 max-w-sm mx-auto">
+    <a onclick="window.history.back()" class="flex-1 p-4 rounded-md bg-black text-white font-semibold hover:bg-black/80">Go Back</a>
```
daniellemaxwell, line 10: I thought that adding a back button could be helpful, but if a user doesn't have a page to go back to we don't currently have a fallback option (ex. login page) so this can be removed.

bbengfort: It's a good thought - I think the only reason a user would get to this page is if they clicked on a link in an email; so probably they wouldn't be able to make too much use of a back button.

What about a "learn more" link/button instead? E.g. provide two choices: learn more or join TRISA network

bbengfort (approved):

pkg/web/templates/sunrise_404.html
daniellemaxwell: I was thinking about that. I'll add a learn more link.

PR #238, opened 2024-11-05
Sunrise Enabled Config and Middleware

bbengfort:
### Scope of changes

Creates a configuration to enable/disable Sunrise and a middleware to enforce that configuration on sunrise-specific routes. If Sunrise is disabled the server should return a 404. This will necessitate changing some existing endpoints, however.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Are the changes to the routes acceptable? Is this sufficient to enable/disable sunrise? 

### Author checklist

### Reviewer(s) checklist

- [ ] The new routes are acceptable, I have created any stories necessary to make changes on the front end

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -35,6 +35,7 @@ type Config struct {
 	Node          TRISAConfig         `split_words:"true"`
 	DirectorySync DirectorySyncConfig `split_words:"true"`
 	TRP           TRPConfig           `split_words:"true"`
+	Sunrise       SunriseConfig       `split_words:"true"`
 	RegionInfo    RegionInfo          `split_words:"true"`
 	processed     bool
 }
@@ -105,6 +106,10 @@ type TRPConfig struct {
 type TRPIdentityConfig struct {
 	VASPName string `split_words:"true" desc:"specify the name response in a trp identity request"`
 	LEI      string `required:"false" desc:"the lei of your vasp to respond to a trp identity request"`
+}
+
+type SunriseConfig struct {
+	Enabled bool `default:"true" desc:"if false, it will not be possible to send sunrise emails and sunrise endpoints will return a 404"`
 }
 
 // Optional region and deployment information associated with the node.
--- a/pkg/web/middleware.go
+++ b/pkg/web/middleware.go
@@ -1,0 +1,71 @@
+package web
+
+import (
+	"net/http"
+	"strings"
+
+	"github.com/gin-gonic/gin"
+	"github.com/gin-gonic/gin/binding"
+)
+
+// If the API is disabled, this middleware factory function returns a middleware that
+// returns 529 Unavailable on all API requests. If the API is enabled, it returns nil
+// and should not be used as a middleware function.
+func (s *Server) APIEnabled() gin.HandlerFunc {
+	if s.conf.Web.APIEnabled {
+		return nil
+	}
+
+	return func(c *gin.Context) {
+		// If this is an API request and the API is disabled, return unavailable.
+		if IsAPIRequest(c) {
+			c.AbortWithStatus(http.StatusServiceUnavailable)
+			return
+		}
+
+		// Otherwise serve the UI request.
+		c.Next()
+	}
+}
+
+// If the UI is disabled, this middleware factory function returrns a middleware that
+// returns 529 Unvailable on all UI requests. If the UI is enabled, it returns nil.
+func (s *Server) UIEnabled() gin.HandlerFunc {
+	if s.conf.Web.UIEnabled {
+		return nil
+	}
+
+	return func(c *gin.Context) {
+		// If this is not an API request, and the UI is disabled, return unavailable.
+		if !IsAPIRequest(c) {
+			c.AbortWithStatus(http.StatusServiceUnavailable)
+			return
+		}
+
+		// Otherwise serve the API request.
+		c.Next()
+	}
+}
+
+// Determines if the request being handled is an API request by inspecting the request
+// path and Accept header. If the request path starts in /v1 and the Accept header is
+// nil or json, then this function returns true.
+func IsAPIRequest(c *gin.Context) bool {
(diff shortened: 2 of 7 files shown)


pkg/web/middleware.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Determines if the request being handled is an API request by inspecting the request
+// path and Accept header. If the request path starts in /v1 and the Accept header is
+// nil or json, then this function returns true.
+func IsAPIRequest(c *gin.Context) bool {
+	if strings.HasPrefix(c.Request.URL.RequestURI(), "/v1") {
+		return !(c.NegotiateFormat(binding.MIMEJSON, binding.MIMEHTML) == binding.MIMEHTML)
+	}
+	return false
+}
```
bbengfort, line 58: This is moved from web.go to keep all the middleware in the same file.

pkg/web/middleware.go
```diff
@@ -0,0 +1,71 @@
...
+	return func(c *gin.Context) {
+		if !enabled {
+			c.AbortWithStatus(http.StatusNotFound)
+			return
+		}
+
+		// Handle the request if the route is enabled
+		c.Next()
+	}
+}
```
bbengfort, line 71: This is the new middleware that returns 404 if sunrise is not enabled.

pkg/web/routes.go
```diff
@@ -102,18 +102,24 @@ func (s *Server) setupRoutes() (err error) {
 	s.router.GET("/transactions/:id/info", authenticate, s.TransactionsInfo)
 	s.router.GET("/send-envelope", authenticate, s.SendEnvelopeForm)
 	s.router.GET("/send-message", authenticate, s.SendMessageForm)
-	s.router.GET("/verify", s.VerifySunriseUser)
-	s.router.GET("/sunrise-accept", s.SunriseMessagePreview)
```
bbengfort, line 106: These routes will be changed to `/sunrise/verify` and `/sunrise/accept` is that ok?

pkg/config/config.go
```diff
@@ -35,6 +35,7 @@ type Config struct {
 	Node          TRISAConfig         `split_words:"true"`
 	DirectorySync DirectorySyncConfig `split_words:"true"`
 	TRP           TRPConfig           `split_words:"true"`
+	Sunrise       SunriseConfig       `split_words:"true"`
```
daniellemaxwell, line 38: Will the `SunriseConfig` need to be added to the scene so that the FE may check to see if it has been enabled?

pkg/config/config.go
```diff
@@ -107,6 +108,10 @@ type TRPIdentityConfig struct {
 	LEI      string `required:"false" desc:"the lei of your vasp to respond to a trp identity request"`
 }
 
+type SunriseConfig struct {
+	Enabled bool `default:"true" desc:"if false, it will not be possible to send sunrise emails and sunrise endpoints will return a 404"`
```
daniellemaxwell, line 112: Setting the value to `true` by default here feels like the easiest way to ensure that each node has it enabled while turning it off is a simple change to the config. If we get feedback requesting another way to enable/disable the feature when we could reassess.

pkg/web/routes.go
daniellemaxwell: That's fine. Should we change `send-message` to `sunrise/message` to make it a little more clear that it's related to that feature? I wasn't sure what would be preferred when initially creating the routes.

pkg/config/config.go
bbengfort: Good point! I'll add it in this PR.

pkg/web/routes.go
bbengfort: Ah, I didn't realize `send-message` was a sunrise route; yes we should probably change it so that it's enabled/disabled by the middleware. I propose `sunrise/send` is that ok?

bbengfort: Er, or I guess `POST sunrise/send` and `GET sunrise/message`?

daniellemaxwell: That'll work!

pkg/web/routes.go
```diff
@@ -101,19 +101,25 @@ func (s *Server) setupRoutes() (err error) {
...
+	// NOTE: should documentation require authentication?
 	s.router.GET("/v1/docs/openapi.:ext", s.OpenAPI)
 	s.router.GET("/v1/docs", s.APIDocs)
 
+	// Sunrise Routes (can be disabled by the middleware)
+	sunrise := s.router.Group("/sunrise", s.SunriseEnabled())
+	{
+		sunrise.GET("/verify", s.VerifySunriseUser)
+		sunrise.GET("/accept", s.SunriseMessagePreview)
+		sunrise.GET("/message", authenticate, s.SendMessageForm)
```
bbengfort, line 120: @daniellemaxwell moved the send-message route here.

pkg/web/scene/scene.go
bbengfort, line 1: @daniellemaxwell I figured we may want to add arbitrary configuration values to the Scene in the future, so I created a `WithConf` package function that can update internal variables from the configuration. Right now `sunriseEnabled` is the only configuration, but I suspect we'll want others in the future (like `EmailIsAvailable` for example). 

Note that if `WithConf` is not called, then `SunriseEnabled` will not be on the scene.

pkg/web/web.go
```diff
@@ -61,56 +60,10 @@ func New(conf config.Config, store store.Store, network network.Network) (s *Ser
...
-			c.AbortWithStatus(http.StatusServiceUnavailable)
-			return
-		}
-
-		// Otherwise serve the UI request.
-		c.Next()
-	}
-}
+	// Update the scene with the configuration
+	scene.WithConf(&conf)
```
bbengfort, line 64: This is where the scene is getting updated with the configuration; we'll have to remember where this is … I'm sure we're going to tear our hair out looking for it eventually.

PR #278, opened 2025-03-06
Create API Keys

bbengfort:
### Scope of changes

Well this was a nightmare.

Implements the Create API Key modal on the API Key dashboard page. Creating the modal went fine … but figuring out how to get HTMX to submit the request correctly was … horrible. 

Some key points:

1. The HTMX docs say that "Note that if an input value appears more than once the value in the parameters object will be an array, rather than a single value" ([ref](https://htmx.org/events/#htmx:configRequest)). **However**, this is no longer true in HTMX v2 because they are now using FormData for everything ([ref](https://github.com/bigskysoftware/htmx/pull/2226)). This made the code from the original version obsolete because we're using v2 for … reasons.
2. Actually what HTMX uses is a `ProxyForm` ([ref](https://github.com/bigskysoftware/htmx/blob/07d35186fbd6ec3074d86b0b041e1b4869a33f36/src/htmx.js#L4050)) -- and this is what actually gets serialized to JSON ([ref](https://github.com/bigskysoftware/htmx/blob/07d35186fbd6ec3074d86b0b041e1b4869a33f36/src/htmx.js#L4104)).
3. So now we need a json-enc extension that encodes `FormData` -- which I've added to the repo in this PR. It checks to see if there are multiple values for the same key and makes it a `[]string` if so; so that works for Permissions.
4. However, we cannot store arbitrary JSON in the `e.detail.parameters` so if we need that; we'll have to figure out how to deal with that in the form data somehow; either by `try/catch` deserializing all JSON objects or via something more clever.

What a mess.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [x] other (describe)

### Acceptance criteria

This PR will be merged without review.

### Author checklist


pkg/web/static/js/htmx-json-enc.js
bbengfort, line 1: @daniellemaxwell if you're interested here's how I had to fix the issue.

PR #293, opened 2025-04-04
Daybreak backend and import command

chris-okuda:
Added a command `go run ./cmd/envoy daybreak:import --in JSONFILE` which can be used to import Daybreak Counterparties and Contacts into the Daybreak Directory in the database. Currently, it will not update any contacts that already exist, but it will update the other properties from a Counterparty.

Also added a VSCode launch config for the debugger for this command and updated Step 1 in the README because if you run the `generate.sh` script it puts the certs in the working directory rather than `.secret/`.


bbengfort (approved):
:shipit:

PR #297, opened 2025-04-22
Add Daybreak counterparties to UI

chris-okuda:
**_TODO_**: need to resolve the test for `SunriseEnabled` and also resolve the placement/naming for the `DaybreakEnabled` or `ShowDaybreak` option (and possibly create a ticket for making enums more developer friendly?)

### Scope of changes

Add Daybreak counterparties to UI.

Also bugfixed the `daybreak:import` command for an edge case and bugfixed the "daybreak" source in `source.go` to properly output the "daybreak" string. Also fixed a few tests for the new source and other issues.

Also, @bbengfort refactored code from the import command function into the database `store` package so that all of the daybreak-related transactions where we create/update a counterparty and it's contacts are done in single related transactions so that if there is an error with a counterparty or a related contact they won't be orphaned in the database to cause issues later.

### Type of change

- [x] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

No other specific requirements.

### Author checklist

### Reviewer(s) checklist


cmd/envoy/main.go
```diff
@@ -574,46 +574,50 @@ func daybreakImport(c *cli.Context) (err error) {
...
-							continue
-						}
-					}
-				}
 			} else {
 				// create the counterparty and all contacts together
 				if err = db.CreateCounterparty(ctx, modelCounterparty); err != nil {
-					log.Warn().Err(err).Msg(fmt.Sprintf("Error when creating Counterparty: '%s' (ID: '%s')", modelCounterparty.Name, modelCounterparty.ID))
-					continue
+					if errors.Is(err, dberr.ErrAlreadyExists) {
```
chris-okuda, line 595: There was a weird edge case that hit this point, where a "daybreak" counterparty was in the DB but not returned from the "source info" step, so in order to try and fix this, we will try and look up the counterparty by the directory_id (which is unique on the hostname/commonname but as a UUID) and then if it is found and it is in the "daybreak.rotational.io" registered_directory, then we'll go ahead and update it.

cmd/envoy/main.go
```diff
@@ -781,3 +785,34 @@ func valueForColumn(table, column string) interface{} {
 
 	panic(fmt.Errorf("unknown type for %s.%s", table, column))
 }
+
+func UpdateCounterpartyAndContacts(ctx context.Context, modelCounterparty *models.Counterparty, apiCounterparty *api.Counterparty) error {
```
chris-okuda, line 792: I had to do this two times so I made it a helper

cmd/envoy/main.go
```diff
@@ -574,46 +574,53 @@ func daybreakImport(c *cli.Context) (err error) {
 			log.Warn().Msg(fmt.Sprintf("Protocol must be 'sunrise' to import: '%s' (DirID: '%s')", apiCounterparty.Name, apiCounterparty.DirectoryID))
 			continue
 		}
+		if modelCounterparty.RegisteredDirectory.Valid && modelCounterparty.RegisteredDirectory.String != "daybreak.rotational.io" {
```
chris-okuda, line 577: Because I use this string to check something later for the rare edge case, we can ensure it's always the same (the daybreak_elt always uses this string)

docker-compose.yaml
```diff
@@ -132,4 +132,4 @@ secrets:
   server.pem.gz:
     file: ./.secret/server.pem.gz
   counterparty.pem.gz:
-    file: ./.secret/counterparty.pem.gz
```
chris-okuda, line 135: every line of a text document should have a newline, technically, or it's not "valid" in some contexts

pkg/web/scene/scene_test.go
```diff
@@ -66,8 +67,12 @@ func TestNew(t *testing.T) {
 			scene.WithConf(&conf)
 
 			data := scene.New(CreateContext())
+
 			require.Contains(t, data, scene.SunriseEnabled, "expected sunrise enabled to be set with conf")
-			require.True(t, data[scene.SunriseEnabled].(bool), "expected sunrise enabled to default to true")
+			require.False(t, data[scene.SunriseEnabled].(bool), "expected sunrise enabled to default to false because email is unavailable by default")
```
chris-okuda, line 72: so, since email is defaulted to not configured, sunrise should NOT be enabled in this test, as far as I can tell? if I am incorrect please let me know to fix this

pkg/config/config.go
```diff
@@ -127,6 +128,10 @@ type WebhookConfig struct {
 	RequireServerAuth bool   `default:"false" split_words:"true" desc:"if true, the webhook server will require a valid hmac authorization header in webhook responses"`
 }
 
+type DaybreakConfig struct {
+	Enabled bool `default:"false" desc:"if true, the Daybreak counterparties list tab will be shown, otherwise not shown"`
+}
```
bbengfort, line 133: Do you think we'll have additional daybreak configs in the future? If not, we could just make this a `ShowDaybreak` bool in the Web config.

pkg/config/config_test.go
```diff
@@ -130,6 +131,7 @@ func TestConfig(t *testing.T) {
 	require.Equal(t, testEnv["REGION_INFO_COUNTRY"], conf.RegionInfo.Country)
 	require.Equal(t, testEnv["REGION_INFO_CLOUD"], conf.RegionInfo.Cloud)
 	require.Equal(t, testEnv["REGION_INFO_CLUSTER"], conf.RegionInfo.Cluster)
+	require.True(t, conf.Web.Daybreak.Enabled)
```
bbengfort, line 134: Thanks for adding this test!

pkg/enum/source.go
```diff
@@ -69,7 +69,7 @@ func ParseSource(s interface{}) (Source, error) {
 }
 
 func (s Source) String() string {
-	if s > SourceRemote {
+	if s > SourceDaybreak {
```
bbengfort, line 72: Good catch - I wish there was a better way to do this with enums.

pkg/web/api/v1/counterparty.go
```diff
@@ -378,8 +378,8 @@ func (c *Contact) Validate(create bool) (err error) {
 
 func (c *CounterpartyQuery) Validate() (err error) {
 	if c.Source != "" {
-		if ok, _ := enum.CheckSource(c.Source, enum.SourceUnknown, enum.SourceDirectorySync, enum.SourceUserEntry); !ok {
-			err = ValidationError(err, IncorrectField("source", "must be one of gds or user"))
+		if ok, _ := enum.CheckSource(c.Source, enum.SourceUnknown, enum.SourceDirectorySync, enum.SourceUserEntry, enum.SourceDaybreak); !ok {
+			err = ValidationError(err, IncorrectField("source", "must be one of gds, user, or daybreak"))
```
bbengfort, line 382: I guess this was what was giving you trouble with the query!

pkg/web/scene/scene.go
```diff
@@ -265,4 +273,8 @@ func WithConf(conf *config.Config) {
 	// Compute the sunriseEnabled boolean
 	enabled := conf.Sunrise.Enabled && conf.Email.Available()
 	sunriseEnabled = &enabled
+
+	// Compute the daybreakEnabled boolean
+	enabled = conf.Web.Daybreak.Enabled
```
bbengfort, line 278: Based on your comment above, should this be as follows?

```suggestion
	enabled = conf.Web.Daybreak.Enabled && *sunriseEnabled
```

pkg/web/scene/scene_test.go
bbengfort: It appears that it is enabled (I think). The `CreateConf` function called below sets the `testEnv` before creating the config and part of the `testEnv` is:

```
 "TRISA_EMAIL_TESTING":          "true",
```

And I think that enables email from a testing perspective (though I'd have to check on that). 

But then that brings up the question -- why was this test passing before and after you made this change? Something is fishy.

pkg/web/templates/docs/openapi/openapi.json
```diff
@@ -767,7 +767,8 @@
             "enum": [
               "gds",
               "user",
-              "peer"
+              "peer",
+              "daybreak"
```
bbengfort, line 771: Thank you for being so thorough and updating every place the code needs to be updated including tests and docs; I really appreciate it! Hopefully that means you're finding the code easy to follow … but if not; let me know I'm always open to feedback.

pkg/web/templates/docs/openapi/openapi.yaml
bbengfort, line 1: I use a VSCode extension to translate the openapi JSON to YAML and I use Frogment to edit the OpenAPI spec. Just FYI in case you have to do more editing.

cmd/envoy/main.go
bbengfort: If it's a helper it can be `updateCounterpartyAndContacts` -- but we could also add it to the database code using a `Daybreak` interface. I'm going to shortly push a fix that shows how that might work and if you like it we can merge it or if you don't, we can revert to this commit.

bbengfort (commented):
A couple of comments and questions for you -- but honestly I'm very impressed at the thoroughness of your work and how you read code to ensure that everything you're doing is super complete! I'm really stoked about collaborating with you on Envoy!

cmd/envoy/main.go
```diff
@@ -574,46 +574,53 @@ func daybreakImport(c *cli.Context) (err error) {
...
 					log.Warn().Err(err).Msg(fmt.Sprintf("Error when updating Counterparty: '%s' (ID: '%s')", modelCounterparty.Name, modelCounterparty.ID))
 					continue
 				}
-
-				// update or create contacts (not performed in `UpdateCounterparty()`)
-				for _, apiContact := range apiCounterparty.Contacts {
-					// convert the api.Contact to a model.Contact
-					var modelContact *models.Contact
-					if modelContact, err = apiContact.Model(modelCounterparty); err != nil {
-						log.Warn().Err(err).Msg(fmt.Sprintf("Error when converting Contact to a model: '%s'", apiContact.Name))
```
bbengfort, line 593: Note that you can use:

```
log.Warn().Err(err).Msgf("Error when converting Contact to a model: '%s'", apiContact.Name)
```

cmd/envoy/main.go
```diff
@@ -574,46 +574,53 @@ func daybreakImport(c *cli.Context) (err error) {
...
 
 		if modelCounterparty.DirectoryID.Valid {
 			cSrc, ok := srcMap[modelCounterparty.DirectoryID.String]
 			if ok {
 				// counterparty is present in DB, so we update it
 				modelCounterparty.ID = cSrc.ID
-				if err = db.UpdateCounterparty(ctx, modelCounterparty); err != nil {
+				apiCounterparty.ID = cSrc.ID
+				if err = UpdateCounterpartyAndContacts(ctx, modelCounterparty, apiCounterparty); err != nil {
 					log.Warn().Err(err).Msg(fmt.Sprintf("Error when updating Counterparty: '%s' (ID: '%s')", modelCounterparty.Name, modelCounterparty.ID))
```
bbengfort, line 589: I would recommend this code:

```suggestion
					log.Warn().Err(err).Str("directory_id", apiCounterparty.DirectoryID).Str("name", apiCounterparty.Name).Msg("could not convert counterparty to model")
```

When we run this in production, the JSON format will be a bit easier to parse and handle in Google Log Explorer in the console. The human readable version isn't too bad either, it's very similar to the string format you have; without the parentheses (and maybe equal instead of colon)

cmd/envoy/main.go
```diff
@@ -781,3 +788,34 @@ func valueForColumn(table, column string) interface{} {
...
+		// create the contact, updating any already-existing contacts
+		if err = db.CreateContact(ctx, modelContact); err != nil {
+			if errors.Is(err, dberr.ErrAlreadyExists) {
+				//TODO: get the contact's ID and update the contact (this improvement has a ticket)
+				continue
+			} else {
+				log.Warn().Err(err).Msg(fmt.Sprintf("Error when creating Contact: '%s' (CounterpartyID: '%s')", modelContact.Name, modelCounterparty.ID))
+				continue
+			}
+		}
```
bbengfort, line 817: I'm going to use `modelContact.SetContacts` here instead when I push my recommendations; I think that will make the single transaction model a bit easier to read and handle.

cmd/envoy/main.go
```diff
@@ -521,7 +519,7 @@ func daybreakImport(c *cli.Context) (err error) {
 	// Load the JSON file into Counterparty objects
 	var in string
 	if in = c.String("in"); in == "" {
-		return cli.Exit("Must pass argument 'in' as a path to a JSON file with a list of Counterparty objects", 1)
+		return cli.Exit("must pass argument 'in' as a path to a JSON file with a list of Counterparty objects", 1)
```
bbengfort, line 522: My go vet doesn't like errors that start with a capital letter …

cmd/envoy/main.go
```diff
@@ -534,93 +532,114 @@ func daybreakImport(c *cli.Context) (err error) {
 		return cli.Exit(err, 1)
 	}
 
-	// db context
+	// Make sure we have a daybreak database
+	ddb, ok := db.(store.DaybreakStore)
+	if !ok {
+		return cli.Exit("configured database does not support daybreak operations", 2)
+	}
```
bbengfort, line 539: As promised here is the specialized daybreak interface that I suggested I would add for you to review. This allows us to push Daybreak code back to the database so that we can perform operations in a single transaction; this will ensure no partial records get created; e.g. a counterparty is updated but then fails on the first contact creation so it's left stranded with no contact. 

However, if you don't like my updates I'm happy to roll them back to the previous commit.

cmd/envoy/main.go
```diff
@@ -534,93 +532,114 @@ func daybreakImport(c *cli.Context) (err error) {
...
-	for _, cSrc := range srcInfo {
-		if cSrc.DirectoryID.Valid {
-			srcMap[cSrc.DirectoryID.String] = cSrc
-		} else {
-			log.Warn().Msg(fmt.Sprintf("CounterpartySourceInfo was missing a DirectoryId (ID: %s)", cSrc.ID))
-		}
-	}
+	// Begin import; all counterparties and associated contacts that can be created will
+	// be; otherwise the entire counterparty will be skipped.
+	log.Info().Msgf("starting to import %d Daybreak Counterparties...", len(cpartyImports))
```
bbengfort, line 555: Using `log.Info().Msgf` here as suggested earlier.

cmd/envoy/main.go
```diff
@@ -534,93 +532,114 @@ func daybreakImport(c *cli.Context) (err error) {
...
 
-	log.Info().Msg(fmt.Sprintf("Starting to import %d Daybreak Counterparties...", len(cpartyImports)))
+	var (
+		updated int
+		created int
+	)
```
bbengfort, line 560: I broke up `importCnt` into two counters to make it easier to see what was happening.

cmd/envoy/main.go
```diff
@@ -534,93 +532,114 @@ func daybreakImport(c *cli.Context) (err error) {
...
 
-	// Create or update each counterparty
-	var importCnt int = 0
 	for _, apiCounterparty := range cpartyImports {
-		//convert the api.Counterparty to a model.Counterparty
+		// Convert the api.Counterparty to a model.Counterparty
 		var modelCounterparty *models.Counterparty
 		if modelCounterparty, err = apiCounterparty.Model(); err != nil {
-			log.Warn().Err(err).Msg(fmt.Sprintf("Error when converting Counterparty to a model: '%s' (DirID: '%s')", apiCounterparty.Name, apiCounterparty.DirectoryID))
+			log.Warn().Err(err).Str("directory_id", apiCounterparty.DirectoryID).Str("name", apiCounterparty.Name).Msg("could not convert counterparty to model")
```
bbengfort, line 566: Employing the log format method I suggested in my earlier review.

cmd/envoy/main.go
```diff
@@ -534,93 +532,114 @@ func daybreakImport(c *cli.Context) (err error) {
...
-						}
-					}
-				}
-			} else {
-				// create the counterparty and all contacts together
-				if err = db.CreateCounterparty(ctx, modelCounterparty); err != nil {
-					log.Warn().Err(err).Msg(fmt.Sprintf("Error when creating Counterparty: '%s' (ID: '%s')", modelCounterparty.Name, modelCounterparty.ID))
-					continue
-				}
+		if !modelCounterparty.RegisteredDirectory.Valid || modelCounterparty.RegisteredDirectory.String != "daybreak.rotational.io" {
```
bbengfort, line 580: Note that I changed this conditional from:

```
modelCounterparty.RegisteredDirectory.Valid && modelCounterparty.RegisteredDirectory.String != "daybreak.rotational.io"
```

I think the warning should occur if its not valid or its not daybreak -- but let me know if I'm incorrect with the logic.

pkg/store/sqlite/daybreak.go
bbengfort, line 1: Most database code ported from `main.go` but it could certainly use a detailed look by you in case I messed something up.

bbengfort (approved):
Ok, so there were quite a lot of comments and I even pushed my suggestions as a commit. All this was by way of trying to be helpful, but if you'd like to revert back to your prior commit and make changes yourself, that is perfectly fine with me. 

I'm going to go ahead and mark this as :shipit: so that when you review the comments and make your last changes you aren't blocked by me because I don't think I need to review your changes again. However, I do strongly recommend that you review everything and my comments before merging in case I didn't leave it in a good state by accident! 

Thanks for your hard work on this, and I'm looking forward to having Daybreak as part of our 1.0.0 release!

cmd/envoy/main.go
chris-okuda: I like it very much, it makes the transactions safer and it moves the database related logic into the `store` where it makes a lot more sense to be.

cmd/envoy/main.go
chris-okuda: Thanks, I was still in the "text string" log mindset, I'll start using the contextual logging feature in zerolog more!

cmd/envoy/main.go
chris-okuda: Yea I think this makes much more sense, otherwise it could just be missing this field.

pkg/config/config.go
chris-okuda: I guess it depends on if we want to add any features or options later down the road? I'll leave this discussion un-resolved for now.

pkg/enum/source.go
chris-okuda: The only thing I can think of would be to put an "end" enum value and then always add new values inside of it. Something like:

`if s <= SourceEndingVal`

with:

```
const (
	SourceUnknown Source = iota
	SourceDirectorySync
	SourceUserEntry
	SourcePeer
	SourceLocal
	SourceRemote
	SourceDaybreak
        SourceEndingVal // NOTE: always add new Source enum values above this one!
)
var sourceNames [7]string = [...]string{
         // NOTE: always add new Source enum strings BEFORE "endingval"
	"unknown", "gds", "user", "peer", "local", "remote", "daybreak", "endingval",
}
```

This would, of course, require the developers to always put the new values in the right spot, but it would also prevent the developer from having to know a few arcane places to swap values later on. It would be designed overall so the "endingval" value strings would never be exposed outside of the specific enum implementation. I'm pretty new to Go, however, so I don't know if this would be a bad implementation idea.

pkg/web/api/v1/counterparty.go
chris-okuda: Partly yea! 

It turns out that I had not changed the Source enum function `String()` properly, so any Daybreak counterparty that was imported with the "daybreak" source was valid according to the `Valid()` code for that enum because I changed that code, but then the string in the database was "unknown" because the default for `String()` was to return that if the given enum value was out of bounds. That's the "rare edge case" that I ran into, where using the "source info" API didn't return "daybreak" counterparties because there were none, they were all "unknown" sources! So, that's why in the "rare edge case" for the "create" branch of the import command we have to check for the "already in database" error and then try to update that counterparty if it seems like the source is incorrect, by checking the "registered directory".

pkg/web/scene/scene.go
chris-okuda: We had discussed leaving it, in case a customer that didn't want to use Sunrise still wanted to see the counterparties in Daybreak for their internal compliance use, so I left it without the && part

pkg/web/scene/scene_test.go
chris-okuda: Interesting, it was failing for me locally and in this branch, so we should look at this some more before allowing this change I made.

pkg/web/templates/docs/openapi/openapi.json
chris-okuda: Yea, it's pretty easy to move around and understand what's what

chris-okuda (commented):
I responded to a bunch of comments, and resolved a bunch, and left 3 or 4 unresolved to discuss later

pkg/web/scene/scene_test.go
chris-okuda: I figure out that inside this test where it calls `scene.WithConf(&conf)` for some reason on the next line where it calls `data := scene.New(CreateContext())` that in `scene.New` it doesn't seem that the `WithConf` had any effect. When I put a breakpoint in each, I can see that inside `WithConf` it does in fact set `sunriseEnabled` to `true`, however then in `scene.New` it appears that `sunriseEnabled` is `false`.

chris-okuda: I solved this through fixing a bug in the `WithConf` where I improperly reused a value `enabled` because I am not quite used to pointers yet :)

pkg/config/config.go
chris-okuda: I think I will leave this as-is for now, it doesn't hurt much to do it like this other than adding a few extra lines of code, and if we do want to add more features later to the config then this will make it easier on the user later so it doesn't break the config format.

PR #302, opened 2025-04-30
Webhook Authentication

bbengfort:
### Scope of changes

This PR implements an authentication mechanism for webhook requests that uses a method similar to [Hawk](https://mohawk.readthedocs.io/en/latest/usage.html#sending-a-request) and [AWS HMAC-256](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html) authentication. 

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Does this authentication method make sense? Any reason to suspect this won't provide good security? Should we perform server validation as well?

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/cmd/envoy/main.go
+++ b/cmd/envoy/main.go
@@ -91,27 +91,6 @@ func main() {
 			},
 		},
 		{
-			Name:     "regentraveladdresses",
-			Usage:    "regenerate travel addresses for accounts and crypto addresses",
-			Category: "admin",
-			Before:   openDB,
-			Action:   regenerateTravelAddresses,
-			After:    closeDB,
-			Flags: []cli.Flag{
-				&cli.StringFlag{
-					Name:    "endpoint",
-					Aliases: []string{"e"},
-					Usage:   "specify an endpoint to generate the addresses with",
-				},
-				&cli.StringFlag{
-					Name:    "protocol",
-					Aliases: []string{"p"},
-					Usage:   "specify the default protocol for the travel addresses",
-					Value:   "trisa",
-				},
-			},
-		},
-		{
 			Name:     "createuser",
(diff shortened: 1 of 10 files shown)


cmd/envoy/main.go
```diff
@@ -552,6 +572,38 @@ func generateTokenKey(c *cli.Context) (err error) {
...
+		return cli.Exit(err, 1)
+	}
+
+	keyid := ulid.Make()
+	keys := hex.EncodeToString(key)
+
+	if out := c.String("out"); out != "" {
+		data := map[string]string{
+			"key":   keys,
+			"keyID": keys,
```
chris-okuda, line 587: I believe this should be `keyid`

pkg/webhook/auth.go
```diff
@@ -0,0 +1,93 @@
...
+// modification to the data to be signed. If the header is already present, it will
+// not be added again. Headers are stored as lowercase values.
+func (h *HMAC) AddHeader(header, value string) {
+	header = strings.ToLower(header)
+	if slices.Contains(h.headers, header) {
+		return
+	}
+
+	h.headers = append(h.headers, header)
+	h.data = append(h.data, []byte(value)...)
```
chris-okuda, line 54: I learned something new: the `...` in Go!

chris-okuda (changes requested):
Check out the one comment about the `"keyID"` value, otherwise it looks great.

cmd/envoy/main.go
bbengfort: good catch - I've made that change!

PR #305, opened 2025-05-05
Allow CounterpartyID for all transactions using "prepare" and "send-prepared"

chris-okuda:
### Scope of changes

The primary method for looking up a Counterparty for any protocol will be the CounterpartyID, with the travel address and name/email being backups for the currently implemented protocols. If the counterparty found does not match the routing protocol requested, then the user will receive an error.

### Type of change

- [ ] bug fix
- [x] new feature
- [x] documentation
- [ ] other (describe)

### Acceptance criteria

No special requirements.

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/web/routing.go
+++ b/pkg/web/routing.go
@@ -78,15 +78,15 @@ func (s *Server) ResolveCounterparty(c *gin.Context, in *api.Routing) (vasp *mod
 	}
 
 	switch {
-		// If vasp is nil we probably shouldn't have made it this far in the code, but protecting ourselves anyway.
-		case vasp == nil:
-			c.JSON(http.StatusNotFound, api.Error("could not identify counterparty from routing information"))
-			return nil, errors.New("unhandled nil vasp at end of resolve counterparty")
-		case vasp.Protocol != protocol:
-			err := errors.New("could not find counterparty that supports requested protocol")
-			c.JSON(http.StatusNotFound, api.Error(err)
-			return nil, err
-		default:
+	// If vasp is nil we probably shouldn't have made it this far in the code, but protecting ourselves anyway.
+	case vasp == nil:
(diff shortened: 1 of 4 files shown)


pkg/web/routing.go
```diff
@@ -32,51 +32,58 @@ func (s *Server) ResolveCounterparty(c *gin.Context, in *api.Routing) (vasp *mod
...
-		// bug that we need to fix ASAP, hence the panic.
-		panic(fmt.Errorf("unhandled protocol in resolve counterparty: %q", protocol.String()))
+	// If the counterparty protocol matches the request protocol then return it, otherwise return an error code
+	if vasp != nil && protocol == vasp.Protocol {
+		return vasp, nil
+	} else {
+		msg := "the counterparty found for the given ID did not match the requested protocol"
+		c.JSON(http.StatusConflict, api.Error(msg))
+		return nil, errors.New(msg)
 	}
```
bbengfort, line 87: I really like this addition! Just a minor suggestion:

```suggestion
	switch {
		// If vasp is nil we probably shouldn't have made it this far in the code, but protecting ourselves anyway.
		case vasp == nil:
			c.JSON(http.StatusNotFound, api.Error("could not identify counterparty from routing information"))
			return nil, errors.New("unhandled nil vasp at end of resolve counterparty")
		case vasp.Protocol != protocol:
			err := errors.New("could not find counterparty that supports requested protocol")
			c.JSON(http.StatusNotFound, api.Error(err)
			return nil, err
		default:
		return vasp, nil
	}
```

Also if you could add the `errors.New` to errors.go in the web package, that will make it easier to test in the future using `errors.Is`.

bbengfort (approved):
Just one minor suggestion but looks great; and thanks as always for updating the openapi docs as well!

PR #320, opened 2025-05-31
Write tests for and complete `store/mock`

chris-okuda:
### Scope of changes

1) Added an `OnFunction()` function to set callbacks for each of the Store interface functions in the mock store
2) Added tests to ensure the mock store helpers and functionality works (does not cover every `OnFunction` or `Function` interface for the mock store, since they are mostly similar)

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Describe how reviewers can test this change to be sure that it works correctly. Add a checklist if possible.

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/store/mock/store.go
+++ b/pkg/store/mock/store.go
@@ -3,7 +3,6 @@ package mock
 import (
 	"context"
 	"database/sql"
-	"fmt"
 	"testing"
 	"time"
 
@@ -46,25 +45,6 @@ func Open(uri *dsn.DSN) (*Store, error) {
 }
 
 //===========================================================================
-// Store Callback Types
-//===========================================================================
-
-type ErrorFn func() error
-type ContextFn func(context.Context, ...any) (any, error)
-type ListStoreFn func(context.Context, any) (any, error)
-type CreateUpdateStoreFn func(context.Context, any) error
-type RetrieveStoreAnyFn func(context.Context, any) (any, error)
-type RetrieveStoreUUIDFn func(context.Context, uuid.UUID) (any, error)
-type RetrieveStoreULIDFn func(context.Context, ulid.ULID) (any, error)
-type DeleteStoreUUIDFn func(context.Context, uuid.UUID) error
-type DeleteStoreULIDFn func(context.Context, ulid.ULID) error
-type TransactionStateFn func(context.Context, uuid.UUID) (bool, enum.Status, error)
-type ListRetrieveAssocFn func(context.Context, any, any) (any, error)
-type ListRetrieveAssocUUIDFn func(context.Context, uuid.UUID, any) (any, error)
-type ListRetrieveAssocULIDFn func(context.Context, ulid.ULID, any) (any, error)
-type DeleteActionAssocFn func(context.Context, any, any) error
-
-//===========================================================================
 // Mock Helper Methods
 //===========================================================================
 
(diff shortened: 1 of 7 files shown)


bbengfort (approved):
These extensive changes look good and I appreciate all the hard work you did on them! I do have one note, but I don't think it's worth the work of making the changes unless you think so:

Basically, it feels like we have an unnecessary type check because we're using the `callbacks` map. E.g. the following pattern:

```go
type Mock struct {
    callbacks map[string]any
}

func (m *Mock) OnDeleteRecord(fn func(context.Context, uuid.UUID) error)) {
    m.callbacks["DeleteRecord"] = fn
}

func (m *Mock) DeleteRecord(ctx context.Context, id uuid.UUID) error {
    return m.callbacks["DeleteRecord"].(func(context.Context, uuid.UUID error))(ctx, id)
}
```

Contains an unnecessary function type check in `DeleteRecord` -- instead you could do:

```go
type Mock struct {
    OnDeleteRecord func(context.Context, uuid.UUID) error
}

func (m *Mock) DeleteRecord(ctx context.Context, id uuid.UUID) error {
    return m.OnDeleteRecord(ctx, id)
}
```

However, since you did all this work already; I don't know if it's worth refactoring yet again to prevent the type check. I'll leave it up to you.

chris-okuda:
> These extensive changes look good and I appreciate all the hard work you did on them! I do have one note, but I don't think it's worth the work of making the changes unless you think so:
> 
> <snipped>
> 
> However, since you did all this work already; I don't know if it's worth refactoring yet again to prevent the type check. I'll leave it up to you.

It wasn't too hard to do this with VSCode's search and multi-cursor feature. I just searched for the regex `fn.\(func\(.*\)\(` which allowed me to replace the function type assertion with the `s.OnFunction` method call in every place it was located, and then did some other simple search and multi-cursor magic to create all of the stubs for `OnFunction` in the mock `Store`. I'll push those changes and verify with you then merge it!

pkg/store/mock/store.go
```diff
@@ -19,11 +21,82 @@ import (
...
+// mock store should be used per test. To set a callback for `Store.Function()`, set
+// the `Store.OnFunction` stub to your desired callback function.
 type Store struct {
-	callbacks map[string]any
-	calls     map[string]int
-	readonly  bool
+	calls    map[string]int
+	readonly bool
+
+	OnClose                          func() error
```
chris-okuda, line 30: Anyone adding `Function` to the `Store` struct will have to add the mock store's `OnFunction` field stub here and then add the `Function` call at the bottom of this file using the boilerplate copied from one of the other functions and that's all.

pkg/store/mock/store.go
```diff
@@ -50,13 +122,21 @@ func Open(uri *dsn.DSN) (*Store, error) {
...
-	// Set maps to nil to free up memory
+	// reset the call counts
 	s.calls = nil
-	s.callbacks = nil
-
-	// Create new calls and callbacks maps
 	s.calls = make(map[string]int)
-	s.callbacks = make(map[string]any)
+
+	// reset the callbacks using reflection
```
chris-okuda, line 129: `Reset` will reset all the `OnFunction` callbacks without having to add to this later on if someone adds to `Store`, which lowers the chances of problems later.

chris-okuda (commented):
A few comments

PR #321, opened 2025-06-04
Tests for `store/models`

chris-okuda:
### Scope of changes

Add tests and helpers to `store/models`. All of the model `Params()` functions are tested, as well as most other functions specific to the models (Scan-like functions will be tested in the `store/sql` package along with other db-implementation-related testing). Some functions were not tested because they are not complex enough or are tested elsewhere.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [x] other (describe)

### Acceptance criteria

Please ensure all tests are written that are required. If you have suggestions, please comment with those and a brief description.

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/store/models/models_test.go
+++ b/pkg/store/models/models_test.go
@@ -1,0 +1,83 @@
+package models_test
+
+import (
+	"database/sql"
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+//==========================================================================
+// Tests
+//==========================================================================
+
+//TODO
+
+//==========================================================================
+// Helpers
+//==========================================================================
+
+// Converts a name (like a field or column name) into a common format for comparison.
+// Example: `some_variable`, `someVariable`, and `SomeVariable` will all become
+// `somevariable` so they can be compared.
+func ConvertNameForComparison(name string) string {
+	// trim whitespace
+	name = strings.TrimSpace(name)
+
+	// remove underscores (for snake_case)
+	name = strings.ReplaceAll(name, "_", "")
+
+	// to lowercase (for CamelCase or camelCase)
+	name = strings.ToLower(name)
+
+	return name
+}
+
+// Returns a slice of strings that have the names of all public fields in the
+// given interface. The names will be converted using `ConvertNameForComparison()`.
+func GetPublicFieldNames(s interface{}) (fields []string) {
+	t := reflect.TypeOf(s)
+	if t.Kind() == reflect.Struct {
+		for _, field := range reflect.VisibleFields(t) {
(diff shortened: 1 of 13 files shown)


pkg/store/models/transaction_test.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func TestTransactionNumEnvelopes(t *testing.T) {
+	// test 1: has envelopes
+	number := getSampleTransaction(true, false).NumEnvelopes()
+	require.Equal(t, int64(0), number, fmt.Sprintf("there should be 0 envelopes but there were %d", number))
+
+	//test 2: no envelopes
+	number = getSampleTransaction(true, true).NumEnvelopes()
+	require.Equal(t, int64(1), number, fmt.Sprintf("there should be 1 envelopes but there were %d", number))
```
bbengfort, line 73: minor nit

```suggestion
	// test 1: no envelopes
	number := getSampleTransaction(true, false).NumEnvelopes()
	require.Equal(t, int64(0), number, fmt.Sprintf("there should be 0 envelopes but there were %d", number))

	//test 2: has envelopes
	number = getSampleTransaction(true, true).NumEnvelopes()
	require.Equal(t, int64(1), number, fmt.Sprintf("there should be 1 envelopes but there were %d", number))
```

pkg/store/models/models_test.go
```diff
@@ -0,0 +1,83 @@
...
+		for _, field := range reflect.VisibleFields(t) {
+			if field.IsExported() && !field.Anonymous { // only collect public fields
+				fields = append(fields, ConvertNameForComparison(field.Name))
+			}
+		}
+		return fields
+	} else {
+		panic(fmt.Sprintf("input must be a struct, but the kind is %s", t.Kind().String()))
+	}
+}
```
bbengfort, line 50: Good use of reflection for these tests!

chris-okuda:
> Looks good! If I understand correctly, these tests mostly focus on the `Params` functions - where you're using reflection to determine the names of the publicly exposed fields and ensuring that they are in the Params unless there are exceptions (which I assume is for models that do not expose some parameters like password).
> 
> Did you have a plan for doing `Scan` testing?

Yea, for Scan the easiest/only way is to just load a model into the DB and back out and ensure the fields match where they are supposed to be and to also load model rows into the database and pull them out, ensuring we cover all the different types of Scan function. The combination of the two should give us security that the Scan functions are working as expected. I was planning to test those in the `store/sql` package but perhaps they will go in here in another PR.

PR #322, opened 2025-06-06
Tests for scanning models

chris-okuda:
### Scope of changes

Add a MockScanner and tests for the Scan functions for all models.

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

No special requirements.

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/store/mock/scanner.go
+++ b/pkg/store/mock/scanner.go
@@ -1,0 +1,83 @@
+package mock
+
+import (
+	"database/sql"
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// MockScanner allows testing `Scan` interfaces for models. You can add an error
+// to return or the data items you wish to scan into a model using `SetError()`
+// or `SetData()`.
+type MockScanner struct {
+	err        error
+	data       []any
+	scanned    int
+	notScanned int
+}
+
+// Scan will return an error if set, otherwise it will check that the lengths of
+// `dest` and the `data` are equal and then attempt to scan each `data` item into
+// the `dest` item with the same index, returning any errors.
+func (m *MockScanner) Scan(dest ...any) (err error) {
+	// if the user set an error, return it
+	if m.err != nil {
+		return m.err
+	}
+
+	// ensure each data item can be scanned into the destination it's assigned to
+	for i, dst := range dest {
+		if scanner, ok := dst.(sql.Scanner); ok {
+			if err = scanner.Scan(m.data[i]); err != nil { // FIXME: pointers to base types aren't recognized here
+				return fmt.Errorf("failed to scan data into destination at index %d: %w", i, err)
+			}
+			m.scanned++
+		} else {
+			m.notScanned++
+			fmt.Printf("item of kind %s at index %d is not a Scanner\n", reflect.TypeOf(dst).Kind().String(), i)
+		}
+
+	}
+
+	return nil
+}
+
+// Assert that the expected number of `data` items were scanned successfully.
+func (m *MockScanner) AssertScanned(t testing.TB, expected int) {
+	require.Equal(t, expected, m.scanned, "expected %d scans, got %d", expected, m.scanned)
+}
+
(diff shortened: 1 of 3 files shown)


pkg/store/mock/scanner.go
```diff
@@ -0,0 +1,83 @@
...
+// the `dest` item with the same index, returning any errors.
+func (m *MockScanner) Scan(dest ...any) (err error) {
+	// if the user set an error, return it
+	if m.err != nil {
+		return m.err
+	}
+
+	// ensure each data item can be scanned into the destination it's assigned to
+	for i, dst := range dest {
+		if scanner, ok := dst.(sql.Scanner); ok {
```
chris-okuda, line 33: @bbengfort This is where I'm not sure what to do for base types for the MockScanner, I guess their pointers don't have the Scanner interface implemented, however I tried a few ways to dereference it and couldn't get it to work right. If you look at the test I made in the Account test file using the MockScanner, more of those are scanned in because they are the NullType types. Thanks for any help!

pkg/store/mock/scanner.go
```diff
@@ -0,0 +1,191 @@
...
+			}
+			m.scanned++
+			continue
+		}
+
+		// Check the common cases without reflection.
+		if err = convertAssign(dst, src); err != nil {
+			if errors.Is(err, ErrUnknownType) {
+				m.notScanned++
+				m.logf("unsupported Scan, storing drive.Value type %T into type %T at index %d", src, dst, i)
```
bbengfort, line 56: FYI the string format `%T` will print the type so that you don't have to use `reflect.TypeOf(dst).Kind().String()`

pkg/store/mock/scanner.go
```diff
@@ -0,0 +1,191 @@
...
+	for _, log := range m.testLog {
+		t.Log(log)
+	}
+	m.testLog = nil
+}
+
+// Internal method to append a log message to the test log.
+func (m *MockScanner) logf(format string, args ...any) {
+	m.testLog = append(m.testLog, fmt.Sprintf(format, args...))
+}
```
bbengfort, line 191: I liked your Print statement for debugging; but I wanted to add it to the testing log instead so we could handle it more gracefully; so I added these functions to help out with test logging.

pkg/store/mock/scanner.go
```diff
@@ -0,0 +1,191 @@
...
+	m.data = data
+}
+
+// Resets the MockScanner to it's original state.
+func (m *MockScanner) Reset() {
+	m.err = nil
+	m.data = nil
+	m.scanned = 0
+	m.notScanned = 0
+	m.testLog = nil
```
bbengfort, line 177: Added these FYI

pkg/store/mock/scanner.go
```diff
@@ -0,0 +1,191 @@
...
+			}
+			*d = s
+			return nil
+		case *[]byte:
+			if d == nil {
+				return ErrNilPtr
+			}
+			*d = []byte(s)
+			return nil
+		}
```
bbengfort, line 84: Basically the pattern here is first check the source type - e.g. what value is in the mock. Then, once we know that; check the destination type -- it _should_ be a pointer to the source type, at which point we can just assign the pointer the value, e.g. `*d = s`; however, we do have to do some nil checking; e.g. a string can't be nil (but this is a test developer error) and also, just because I'm extra; I added some conversion, like []byte to string -- which is probably not necessary for tests.

pkg/store/mock/scanner.go
```diff
@@ -0,0 +1,191 @@
...
+		case *time.Time:
+			*d = s
+			return nil
+		case *string:
+			if d == nil {
+				return ErrNilPtr
+			}
+			*d = s.Format(time.RFC3339Nano)
+			return nil
+		}
```
bbengfort, line 117: Honestly, I meant to do this the other way around; e.g. I could see a tester wanting to add "2025-05-23T12:34:32.21233Z" to the mock instead of a `time.Time` but I don't know why they would add a `time.Time` to a field that's a string … so that's my bad!

PR #329, opened 2025-07-02
Modify the compliance audit log model

chris-okuda:
### Scope of changes

Modify the ComplianceAuditLog model based on feedback and new information.

* Rename ResourceActionMeta to ChangeNotes
* Add an Algorithm field for the signature algorithm
* Modify the ListCAL function to return everything except the change notes or signature-related values (just the summary info)
* Add a RetrieveCAL function to retrieve an entire log with the change notes and signature-related values
* Update tests and test data

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [x] other (describe)

### Acceptance criteria

Describe how reviewers can test this change to be sure that it works correctly. Add a checklist if possible.

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/store/models/audit.go
+++ b/pkg/store/models/audit.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/trisacrypto/envoy/pkg/enum"
+	"github.com/trisacrypto/envoy/pkg/store/errors"
 	"go.rtnl.ai/ulid"
 )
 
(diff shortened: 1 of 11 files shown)


pkg/store/models/audit.go
```diff
@@ -32,27 +33,57 @@ type ComplianceAuditLog struct {
 	ResourceModified time.Time
 	// Action is the type of change made in the database
 	Action enum.Action
-	// ResourceActionMeta is an optional string specific to the ResourceType and
-	// Action that can include further details, such as a JSON changeset or a note
-	ResourceActionMeta sql.NullString
+	// ChangeNotes is an optional string that can include further details
+	ChangeNotes sql.NullString
```
chris-okuda, line 37: I didn't like the previous name, so I renamed it now before we'd have to make a new migration file.

pkg/store/models/audit.go
```diff
@@ -32,27 +33,57 @@ type ComplianceAuditLog struct {
...
+
 	// Signature is a cryptographic Signature that can be used to verify that an
 	// instance of a ComplianceAuditLog was not modified
 	Signature []byte
 	// KeyID is the identification for the public key that can verify this log
 	KeyID string
+	// Algorithm is the identification for the algorithm that can verify this log
+	Algorithm string
```
chris-okuda, line 45: I added this to hold the algo string like `"RSA-PSS-SHA512"`

pkg/store/models/audit.go
```diff
@@ -32,27 +33,57 @@ type ComplianceAuditLog struct {
...
 }
 
-// Adds a signature value to the ComplianceAuditLog, replacing any value present.
+// ###########################################################################
```
chris-okuda, line 48: This section is for the future and was only included here by accident (it hurts nothing at this time so I won't remove it for now.)

pkg/store/sqlite/audit.go
```diff
@@ -50,14 +50,31 @@ func (s *Store) CreateComplianceAuditLog(ctx context.Context, log *models.Compli
 	return nil
 }
 
+func (s *Store) RetrieveComplianceAuditLog(ctx context.Context, id ulid.ULID) (log *models.ComplianceAuditLog, err error) {
```
chris-okuda, line 53: We'll now be providing just the "summary" info for a log rather than everything now in List, so I added Retrieve.

pkg/store/sqlite/testdata/compliance_audit_logs.sql
```diff
@@ -1,11 +1,17 @@
...
-    ('01JZ1HNFJ96A43DZCFYER9NP70', '01HWQE29RW1S1D8ZN58M528A1M', 'user', '01JXTQCDE6ZES5MPXNW7K19QVQ', 'counterparty', '2024-04-04T15:30:35.153035-10:00', 'create', 'test_user_create_counterparty', x'1234567890abcdef', 'abcdef1234567890'),
-    ('01JZ1HNFJ9GRFCV135M3KWYEBH', '01HWQEJJDMS5EKNARHPJEDMHA4', 'api_key', '01JXTGSFRC88HAY8V173976Z9D', 'sunrise', '2024-05-05T16:40:45.164045-10:00', 'delete', 'test_api_key_delete_sunrise', x'1234567890abcdef', 'abcdef1234567890'),
-    ('01JZ1HNFJ91E1A0HFFZ3XBWKAF', '01JXTGSFRC88HAY8V173976Z9D', 'sunrise', '2c891c75-14fa-4c71-aa07-6405b98db7a3', 'transaction', '2024-06-06T17:50:55.175055-10:00', 'update', 'test_sunrise_update_transaction', x'1234567890abcdef', 'abcdef1234567890')
+INSERT INTO compliance_audit_log (id, actor_id, actor_type, resource_id, resource_type, resource_modified, action, change_notes, signature, key_id, algorithm) VALUES
+    -- ID: "01JZ1HNFJ9KTA3Z6Q4RB3X9W2T"
+    (x'0197c31abe499e943f9ae4c2c7d4f05a', '01HWQE29RW1S1D8ZN58M528A1M', 'user', '01HV6QS6AK4KNS46Q9HEB7DTPR', 'account', '2024-01-01T12:00:05.120005-10:00', 'create', 'test_user_create_account', x'1234567890abcdef', 'abcdef1234567890', 'MOCK'),
```
chris-okuda, line 6: @bbengfort I found out why IDs are normally in hex: because the ULID package implements the storage driver Value() function using bytes from MarshalBinary, so when testing RetrieveCAL it wouldn't work if this value was a string with the actual ID, even if the ULID will properly scan in either bytes or a string. Basically, it wouldn't match in it's string form, even if the ListCAL would return them with a proper ULID!

PR #340, opened 2025-08-15
Audit log "missing foreign ID ref" error fixes

chris-okuda:
### Scope of changes

Audit logs require some information to be added to the `context.Context` object which we use to perform `Store` operations. For any operations that a user initiates, such as from the UI or with an API key or a Sunrise message, the actor information will be set properly. For operations such as CLI operations or internal server operations, such as directory sync, there was no actor set in the context to use to create audit logs. Also, for some CLI operations the audit log was not initialized with the server's configured certificates for signing purposes.

This PR fixes the CLI command issues and the directory sync issue, however we may still have places where there are issues wherever the user does not initiate an operation, so I also set the default actor context to "unknown" within the BeginTx function so that transactions do not fail due to audit logs and they are still created (there is also a Warn log with that default setting).

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Please indicate if there are other operations which change the database which are not initiated by a user or API key that I would need to update to work in a similar way to the directory sync or the CLI commands.

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/cmd/envoy/main.go
+++ b/cmd/envoy/main.go
@@ -394,6 +394,11 @@ func createUser(c *cli.Context) (err error) {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
+	// Setup the audit log
+	if ctx, err = setupAuditLog([]byte("createUser"), ctx); err != nil {
+		return cli.Exit(err, 1)
+	}
+
 	if err = db.CreateUser(ctx, user, &models.ComplianceAuditLog{
 		ChangeNotes: sql.NullString{Valid: true, String: "createUser()"},
 	}); err != nil {
@@ -456,6 +461,11 @@ func createAPIKey(c *cli.Context) (err error) {
 
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
+
+	// Setup the audit log
+	if ctx, err = setupAuditLog([]byte("createAPIKey"), ctx); err != nil {
+		return cli.Exit(err, 1)
+	}
 
 	if err = db.CreateAPIKey(ctx, key, &models.ComplianceAuditLog{
 		ChangeNotes: sql.NullString{Valid: true, String: "createAPIKey()"},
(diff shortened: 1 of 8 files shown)


cmd/envoy/main.go
```diff
@@ -755,6 +741,28 @@ func daybreakRetire(c *cli.Context) (err error) {
 // Helper Functions
 //===========================================================================
 
+// Setup the context with the given actor ID (usually the function or command
+// name) and sets up the keychain to sign audit logs.
+func setupAuditLog(actorId string, ctx context.Context) (newCtx context.Context, err error) {
```
bbengfort, line 746: Should the actorID be `"cli" or `"system"``? I think of actor as the thing that's initiating the action, e.g. the user ID of the logged in user - or is it the function name in other places as well?

pkg/enum/actor.go
```diff
@@ -25,12 +26,13 @@ const (
...
 )
 
-var actorNames = [5]string{
+var actorNames = [6]string{
 	"unknown",
 	"user",
 	"api_key",
 	"sunrise",
 	"cli",
+	"internal",
```
bbengfort, line 35: Could we make this `"system"` instead please?

cmd/envoy/main.go
bbengfort: Ok, I do see there is an `ActorCLI` = so what is the ID?

pkg/store/sqlite/store.go
```diff
@@ -102,11 +102,19 @@ func (s *Store) BeginTx(ctx context.Context, opts *sql.TxOptions) (_ *Tx, err er
...
-	actorType, _ := audit.ActorType(ctx)
+	// Get the actor's ID and type. By default, use an "unknown" actor so that
+	// database transactions do not fail for this reason.
+	var (
+		actorID   []byte
+		actorType enum.Actor
+		ok        bool
+	)
+	if actorID, ok = audit.ActorID(ctx); !ok {
+		actorID = []byte("unknown")
```
chris-okuda, line 113: @bbengfort Now we'll have an "unknown" actor setup by default. This would allow any transaction in the DB to be completed even if the actor context was not properly setup. We will still want to find all of the areas this affects and setup an actor context, so we'll have to discuss where other system operations occur.

bbengfort (approved):
Ok, using `unknown` for now makes sense to me; because we'll be able to query on those audit logs and make fixes where necessary. Maybe we can add a `log.Warn()` -- unknown actor so that we can catch this happening in the logs and make fixes as needed?

chris-okuda:
> Ok, using `unknown` for now makes sense to me; because we'll be able to query on those audit logs and make fixes where necessary. Maybe we can add a `log.Warn()` -- unknown actor so that we can catch this happening in the logs and make fixes as needed?

I will add a logging mechanism for that for sure, another thing is when the filtering works on the UI we'll be able to find those as well. Also, I could make the default as "system" rather than "unknown", because I am pretty sure that all of the user-initiated db changes are covered already, so the only ones that might not be covered are "system" operations like receiving a TRISA message.

PR #342, opened 2025-08-28
Audit Log API docs

chris-okuda:
### Scope of changes

Added the audit log endpoints and data formats to the OpenAPI JSON and then converted the JSON to YAML using the `yq` command line tool (`yq -p json -o yaml openapi.json > openapi.yaml`).

Also fixed a bug where the DetailedLogs query param was not being read as a url query param.

NOTE: I used Stoplight, so the entries may look different than before when @bbengfort was using other tools.

### Type of change

- [ ] bug fix
- [ ] new feature
- [x] documentation
- [ ] other (describe)

### Acceptance criteria

Note: The number of spaces for the JSON formatting is different when using the Stoplight tool and it also adds some metadata, so it may appear as if every line has changed in the diff, when in reality only the audit logs details and some "x-whatever" keys were added. 

If we don't want to have the "x-whatever" keys then I can try to remove them using a quick script or something, but we'll have to do that each time.

### Author checklist

### Reviewer(s) checklist


pkg/web/api/v1/audit.go
```diff
@@ -119,7 +119,7 @@ type ComplianceAuditLogQuery struct {
 
 	// DetailedLogs will return the full log entry if true (otherwise List
 	// returns only summary info)
-	DetailedLogs bool `json:"detailed_logs,omitempty"`
+	DetailedLogs bool `json:"detailed_logs,omitempty" form:"detailed_logs" url:"detailed_logs,omitempty"`
```
daniellemaxwell, line 122: Nice fix! It took me forever to realize that not including `form` was the cause of a bug I encountered not long after I first started.

daniellemaxwell (approved):
Thanks for explaining that Stoplight added metadata to the documentation. If it's not going to cause an issue for users when viewing the docs, then I think that it will not be a problem. It could be helpful to mention in the README that the docs were generated using Stoplight and how the metadata shouldn't be updated manually. In addition, we could include the script for updating the docs. You can decide what you feel will be best.

